PR #117, opened 2014-08-13
Multilingual: Refactor link output into module

linc:
Instead of just barfing links in the Foot asset, this refactors the output into a module that can be placed via Smarty & creates link classes for doing CSS image substitutions. Also minimally tweak the default CSS to not be mashed at the bottom of the page, and removes the ridiculous English "Choose language" instruction.


linc:
Resolved conflicts. Would be great to get in this week for the merge.

PR #143, opened 2014-10-20
AuthorSelector: Fix potential jQuery error.

linc:


linc:
@beckyvb Awesome, thank you!

PR #169, opened 2014-12-14
FileUpload: fix backwards compat (PermissionCategory)

linc:
Grabbing peregrine's fix in for review before it hits master.

FileUpload's only hurdle to 2.1 compatibility is the use of CategoryModel:PermissionCategory so we're looking at a workaround of detecting its absence and using a fork of it.


linc:
I don't like this code duplication. I'd rather see the method backported to 2.1 core, if possible.

PR #173, opened 2015-01-02
Signatures plugin: Fix regex for matching bbcode image tags.

beckyvb:
Fixes bug where signatures could have more signatures than allowed if no spaces were between the multiple image tags.

### DIFF
--- a/plugins/Signatures/class.signatures.plugin.php
+++ b/plugins/Signatures/class.signatures.plugin.php
@@ -267,7 +267,7 @@ class SignaturesPlugin extends Gdn_Plugin {
    public function CheckNumberOfImages($Values, &$Sender) {
       if (C('Plugins.Signatures.MaxNumberImages') && C('Plugins.Signatures.MaxNumberImages') !== 'Unlimited') {
          $max = C('Plugins.Signatures.MaxNumberImages');
-         $Sig = Gdn_Format::To($Values['Plugin.Signatures.Sig'], $Sender->Form->GetFormValue('Format'));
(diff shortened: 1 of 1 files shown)


linc:
Why is the version jumping from 1.5.6 to 1.6.1?

beckyvb:
It's currently on version 1.6.0 on stage. I wasn't sure how to deal with the discrepancy. Agree that hopping from 1.5.6 to 1.6.1 does seem somewhat disingenuous.

linc:
Master is the canonical version, so I think 1.5.7 would be correct.

tburry:
This really should not be checking against BBCode. It should be formatting whatever signature into HTML and then checking the number of images there. Otherwise it will fail for Markdown or whatever other format we end up supporting.

plugins/Signatures/class.signatures.plugin.php
```diff
@@ -267,7 +267,8 @@ public function CheckSignatureLength($Values, &$Sender) {
    public function CheckNumberOfImages($Values, &$Sender) {
       if (C('Plugins.Signatures.MaxNumberImages') && C('Plugins.Signatures.MaxNumberImages') !== 'Unlimited') {
          $max = C('Plugins.Signatures.MaxNumberImages');
-         $numMatches = preg_match_all('/(<img|\[img.*\]|\!\[.*\])/i', $Values['Plugin.Signatures.Sig']);
+         $Sig = Gdn_Format::To($Values['Plugin.Signatures.Sig'], $Sender->Form->GetFormValue('Format'));
```
tburry: This is odd code. If the format isn't coming in `$Values` then it should. You also need to default the format to `C('Garden.InputFormatter')`.

tburry: Also you can't wrong with `val('Plugin.Signatures.Sig', $Values)`.

PR #182, opened 2015-02-04
Use SaveImageAs() for images

webstey:
As per issue #2345 , Handle EXIF orientation in UploadImage. Must be accompanied by PR #2348, Fix rotation issue with iPhone uploads.

### DIFF
--- a/plugins/FileUpload/class.fileupload.plugin.php
+++ b/plugins/FileUpload/class.fileupload.plugin.php
@@ -614,7 +614,7 @@ class FileUploadPlugin extends Gdn_Plugin {
       if (is_numeric($MediaID)) {
          // Save the thumbnail information.
          $Model = new MediaModel();
-         $Media = array('MediaID' => $MediaID, 'ThumbWidth' => $Width, 'ThumbHeight' => $Height, 'ThumbPath' => $ThumbParsed['SaveName']);
+         $Media = array('MediaID' => $MediaID, 'ThumbWidth' => $ThumbParsed['Width'], 'ThumbHeight' => $ThumbParsed['Height'], 'ThumbPath' => $ThumbParsed['SaveName']);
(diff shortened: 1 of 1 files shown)


webstey:
Related: #2438

webstey:
https://github.com/vanilla/vanilla/issues/2345

webstey:
Related: https://github.com/vanilla/vanilla/pull/2438

linc:
Perfect! I may ask for some help getting Advanced Editor working with this change as well. Unfortunately it suffers the same problem. Hopefully it will be public shortly.

webstey:
Awesome @linc !  Having advanced editor will be well worth any fixes I can contribute!

plugins/FileUpload/class.fileupload.plugin.php
```diff
@@ -827,7 +827,11 @@ public function PostController_Upload_Create($Sender) {
          $this->EventArguments['OriginalFilename'] = $FileName;
          $Handled = FALSE;
          $this->EventArguments['Handled'] =& $Handled;
-         $this->FireAs('Gdn_Upload')->FireEvent('SaveAs');
+         if($ImageType !== FALSE) {
+            $this->FireAs('Gdn_Upload')->FireEvent('SaveImageAs');
```
linc: What was the purpose behind throwing a differently-named event here? Typically we'd want to do something like add an event argument for `$ImageType` rather than throw two different events.

linc: See https://github.com/vanilla/addons/pull/189 - had to modify this.

webstey: Not sure ... There's a good chance that at the time I thought I was fitting
in with how I thought it works. Thanks for catching that!

PR #183, opened 2015-02-04
QnA: Confirm announcements are a dataset

linc:
Fixes potential fatal error when no announcments were present.


beckyvb:
Looks good!

PR #185, opened 2015-02-05
Redirector: Add legacy redirects, fix for multiple URL parameters named p

initvector:
These updates are meant to be used with the latest "Legacy Slug" updates to ForumMerge.  When a forum is merged in with these updates, you specify a slug as a kind of namespace for the old incoming forum identifiers.  This slug is then specified as the value for the legacy URL parameter.

Redirector checks to see if legacy is set in the URL.  If it is, it checks to see if ForeignID exists in the table (this column is new to Category and Comment, added by ForumMerge).  If both of those checkout, it concatenates the current forum record (category, comment or discussion) identifier with the legacy slug and searches for a row in the target table with that ForeignID.  The first match that is found is returned as the target record.

Also included is a small fix for multiple p parameters being specified in the URL.  When PHP parses the query string into the $_GET array, it only keeps the last value for non-array parameters with duplicate names.  Vanilla uses p to determine its request path and then unsets the value.  This processes causes us to lose both p values and typically causes the wrong p value to be used as the request path.  This is a problem with some other forum URLs the Redirector plug-in handles, because they use a p parameter.  For example, showpost.php?p=100 becomes index.php?p=showpost.php&p=100.  Our path in Vanilla should be set to showpost.php, but it becomes 100.  To compensate for this, the QUERY_STRING server variable is parsed and the duplicate p values are assigned with the assumption that the first p Vanilla's request path and the second p is intended to be a parameter in other forum's URL scheme.  This fix is currently restricted to requests with one p parameter set to showpost.php, showthread.php or viewtopic.php (URLs with the potential to have their own p parameter).

Example usage:
Multiple vBulletin forums are imported into Vanilla.  These new Vanilla forums are merged together, each with their own unique "Legacy Slug" given at the time of merge.  The Redirector plug-in is enabled on the base Vanilla forum. The original vBulletin sites are configured to forward their URLs to the base Vanilla forum site and appending &legacy=(legacy slug) to the end of each URL.  These URLs will trigger a redirect to the new record's location on the Vanilla forum installation.

Comment Example:
* Legacy Slug: vba
* vBulletin Original Comment ID: 2
* Vanilla Comment ID After Merge: 8
* vBulletin Original URL: showpost.php?p=2
* Vanilla Redirector URL: showpost.php?p=2&legacy=vba (index.php?p=showpost.php&p=2&legacy=vba)
* Destination URL: index.php?p=/discussion/comment/8#Comment_8

Additional vBulletin URL Examples:
* Discussion: showthread.php?t=(original discussion ID)
* Category: forumdisplay.php?f=(original category ID)

### DIFF
--- a/plugins/QnA/class.qna.plugin.php
+++ b/plugins/QnA/class.qna.plugin.php
@@ -787,14 +787,16 @@ class QnAPlugin extends Gdn_Plugin {
    }
 
    public function DiscussionModel_BeforeGet_Handler($Sender, $Args) {
-      $Unanswered = Gdn::Controller()->ClassName == 'DiscussionsController' && Gdn::Controller()->RequestMethod == 'unanswered';
+      if (Gdn::Controller()) {
+         $Unanswered = Gdn::Controller()->ClassName == 'DiscussionsController' && Gdn::Controller()->RequestMethod == 'unanswered';
 
-      if ($Unanswered) {
-         $Args['Wheres']['Type'] = 'Question';
-         $Sender->SQL->WhereIn('d.QnA', array('Unanswered', 'Rejected'));
-         Gdn::Controller()->Title('Unanswered Questions');
-      } elseif ($QnA = Gdn::Request()->Get('qna')) {
-         $Args['Wheres']['QnA'] = $QnA;
+         if ($Unanswered) {
+            $Args['Wheres']['Type'] = 'Question';
+            $Sender->SQL->WhereIn('d.QnA', array('Unanswered', 'Rejected'));
+            Gdn::Controller()->Title('Unanswered Questions');
+         } elseif ($QnA = Gdn::Request()->Get('qna')) {
+            $Args['Wheres']['QnA'] = $QnA;
+         }
       }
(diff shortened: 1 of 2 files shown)


linc:
Something in my nginx setup is making `/showthread.php?t=1&legacy=legacy` throw a "File not found." server error. I think we're going to need to get this on cluster 100 for testing. First, here's a couple things:

* Does the `t` parameter work? In vBulletin URLs, the `p` is always for post (comment) and `t` is for thread (discussion), so it's important the Redirector (and this mod) accept either.
* Test this with Q&A enabled. I got a stack trace (accessing property on a non-object) from that scenario.

initvector:
Q&A plug-in was attempting to access Gdn::Controler() as an object without checking to see if it had been set.  If it isn't set, Gdn::Controller() returns NULL.  I added a simple check to the plug-in.

The `t` parameter works for showthread.php.  `p` should work for showpost.php.  Does the following work on your nginx setup? /index.php?p=showthread.php&t=1&legacy=legacy

PR #186, opened 2015-02-11
Use core spoilers in NBBC plugin.

MightyCoder:
Use the same spoilers classes no matter the plugin.


linc:
I'd really like to see core patching to use T('Spoiler') as the toggle instead of an ellipsis before we do this. The core version of this is worse UI. Discussion: https://github.com/vanilla/internal/pull/114

PR #190, opened 2015-02-24
Adding vB4 "Friendly URL" Handling to Redirector Plug-in

initvector:
"Friendly URLs" were added in vB4.  Example: /showthread.php?001-demo-thread/page5

PHP interprets these as named parameters with no value.  We need to extract the record ID and potentially a page number from that parameter name.  For the example, we need to extract 001 as a discussion ID and 5 as the page number.  That is accomplished with this feature.


linc:
NO DOC BLOCKS NO SOUP FOR YOU

PR #195, opened 2015-03-09
Pockets: Add 'Ad Free' functionality

beckyvb:
This PR adds functionality to hide pockets that are marked as 'ads' for any user with the Garden.AdFree.Allow permission.

Add 'Custom' pocket location
Add permission setting (Garden.AdFree.Allow) 
Add 'Ad' type to pockets
Validate Smarty-rendered pockets with 'Custom' locations (using CanRender function)
Use javascript to hide Repeat options when 'Custom' location is set.

### DIFF
--- a/plugins/Pockets/class.pockets.plugin.php
+++ b/plugins/Pockets/class.pockets.plugin.php
@@ -17,7 +17,7 @@ $PluginInfo['Pockets'] = array(
    'AuthorEmail' => 'todd@vanillaforums.com',
    'AuthorUrl' => 'http://vanillaforums.org/profile/todd',
    'RequiredApplications' => array('Vanilla' => '2.1'),
-   'RegisterPermissions' => array('Plugins.Pockets.Manage'),
+   'RegisterPermissions' => array('Plugins.Pockets.Manage', 'Garden.AdFree.Allow'),
    'SettingsUrl' => '/settings/pockets',
    'SettingsPermission' => 'Plugins.Pockets.Manage',
    'MobileFriendly' => TRUE,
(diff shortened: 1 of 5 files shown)


beckyvb:
Relies on https://github.com/vanilla/vanilla/pull/2558

plugins/Pockets/library/class.pocket.php
```diff
@@ -53,9 +56,12 @@ class Pocket {
    /** $var array The repeat frequency. */
    public $MobileNever = FALSE;
 
+   /** $var boolean Is the pocket an ad? */
+   public $IsAd = FALSE;
```
tburry: The property should be the type and there should be an `IsAd()` convenience method.

plugins/Pockets/class.pockets.plugin.php
```diff
@@ -391,7 +398,14 @@ public static function PocketString($Name, $Data = NULL) {
 
       $Result = '';
       foreach ($Pockets as $Pocket) {
-         $Result .= $Pocket->ToString();
+         if (val('Location', $Pocket) == 'Custom' ) {
+            $Data['PageName'] = Pocket::PageName($Data['sender']->Controller->ControllerName);
```
tburry: Use `Gdn::Controller()` and assign a variable before the loop.

plugins/Pockets/library/class.pocket.php
```diff
@@ -22,6 +22,9 @@ class Pocket {
    const REPEAT_EVERY = 'every';
    const REPEAT_INDEX = 'index';
 
+   const AD_TYPE = 'ad';
```
tburry: Constants are almost always in the form `TYPE_*` rather than `*_TYPE`.

PR #196, opened 2015-03-12
QnA: Get count of unanswered discussions to add to pager.

beckyvb:
Also:
Add index to structure.
Add separate method to get unanswered count.

Fixes bug where unanswered pager gets count from discussions.
Relies on https://github.com/vanilla/vanilla/pull/2565


tburry:
Just bump the plugin version and I'll merge.

PR #201, opened 2015-03-20
Update the mulitlingual plugin to work with locale canonicalization

tburry:
This update is in support of the locale PR [here](https://github.com/vanilla/vanilla/pull/2586).

- Improve the locale listing functionality.
- Include the default locale in a less hacky way.
- Use ext-intl for locale names where possible.


linc:
If an old code like `vf_de_DE` is in your config, it will now appear off in the Dashboard and not work, but it will display as an option in Multilingual.

PR #202, opened 2015-03-24
Resolved: de-conflict with blog comments

linc:
Blog comments showed as unresolved, and resolving them closed them.


linc:
Ready for another look.

beckyvb:
Tested and good to go!

PR #207, opened 2015-04-21
Let TouchIcon use updated paths

linc:
Fixes Touch Icons not being able to be uploaded on cloud with CDN.

### DIFF
--- a/plugins/TouchIcon/class.touchicon.plugin.php
+++ b/plugins/TouchIcon/class.touchicon.plugin.php
@@ -95,7 +95,8 @@ class TouchIconPlugin extends Gdn_Plugin {
     * @return string Path to icon
     */
    public function getIconPath() {
-      return Gdn_Upload::Url(C('Garden.TouchIcon', self::DEFAULT_PATH));
+      $Icon = C('Garden.TouchIcon') ? Gdn_Upload::Url(C('Garden.TouchIcon')) : self::DEFAULT_PATH;
+      return $Icon;
    }
 
    /**
--- a/plugins/TouchIcon/class.touchicon.plugin.php
+++ b/plugins/TouchIcon/class.touchicon.plugin.php
@@ -25,14 +25,14 @@ class TouchIconPlugin extends Gdn_Plugin {
          'Internal'
       );
    }
-   
+
    /**
     * Remove route.
     */
    public function OnDisable() {
       Gdn::Router()->DeleteRoute('apple-touch-icon.png');
    }
-   
+
    /**
     * Touch icon management screen.
     *
@@ -43,7 +43,7 @@ class TouchIconPlugin extends Gdn_Plugin {
       $Sender->Permission('Garden.Settings.Manage');
       $Sender->AddSideMenu('dashboard/settings/touchicon');
       $Sender->Title(T('Touch Icon'));
-      
+
       if ($Sender->Form->AuthenticatedPostBack()) {
          $Upload = new Gdn_UploadImage();
          try {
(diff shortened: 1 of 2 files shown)


tburry:
This is not correct.

* You must generate a new random filename when uploading a new image or else CDN cache issues will arise.
* I really want stuff like this to save into a subfolder such as `touch-icons`.

linc:
Changes:

* Use standard-format config name (matching Banner settings)
* Use 'banner' subfolder (where I'd like to move the rest of the Banner images ultimately too).
* Add backward compatibility with current setup via Structure.

Ultimately I'd like this to be on the Banner page with the rest of the settings, hence some of the choices here.

beckyvb:
Looks good to me!

plugins/TouchIcon/class.touchicon.plugin.php
```diff
@@ -43,33 +60,45 @@ public function SettingsController_TouchIcon_Create($Sender) {
...
+      $Sender->Render($this->GetView('touchicon.php'));
+   }
+
+   /**
+    * Get path to icon.
+    *
+    * @return string Path to icon
+    */
+   public function getIconPath() {
+      $Icon = C('Garden.TouchIcon') ? Gdn_Upload::Url(C('Garden.TouchIcon')) : self::DEFAULT_PATH;
```
tburry: You have one case returning a url and the other returning a path. Won't that fail in subfolders?

plugins/TouchIcon/class.touchicon.plugin.php
```diff
@@ -77,10 +106,8 @@ public function SettingsController_TouchIcon_Create($Sender) {
     * @access public
     */
    public function UtilityController_ShowTouchIcon_Create($Sender) {
-      $DefaultPath = 'plugins/TouchIcon/design/default.png';
-      $IconPath = Gdn_Upload::Url('TouchIcon/apple-touch-icon.png');
-      $Redirect = (C('Plugins.TouchIcon.Uploaded')) ? $IconPath : $DefaultPath;
+      $Redirect = $this->getIconPath();
```
tburry: Usually endpoints like this need to be 302s.

plugins/TouchIcon/class.touchicon.plugin.php
```diff
@@ -25,14 +31,25 @@ public function Setup() {
...
    }
-   
+
+   /**
+    * Updates.
+    */
+   public function Structure() {
+      // Backwards compatibility with v1.
+      if (C('Plugins.TouchIcon.Uploaded')) {
+         SaveToConfig('Garden.TouchIcon', PATH_ROOT.'/uploads/TouchIcon/apple-touch-icon.png');
```
tburry: This seems wrong. You are mixing an absolute path with an uploaded path.

plugins/TouchIcon/class.touchicon.plugin.php
```diff
@@ -43,33 +60,45 @@ public function SettingsController_TouchIcon_Create($Sender) {
...
                $Upload->SaveImageAs(
                   $TmpImage,
-                  PATH_ROOT.'/uploads/TouchIcon/apple-touch-icon.png',
+                  $TouchIconPath,
                   114,
                   114,
                   array('OutputType' => 'png', 'ImageQuality' => '8')
                );
-               SaveToConfig('Plugins.TouchIcon.Uploaded', TRUE);
+               SaveToConfig('Garden.TouchIcon', $TouchIconPath);
```
tburry: The `$Upload->SaveImageAs()` method returns an array with a `SavePath` key (I think). That should be used in the config instead.

plugins/TouchIcon/class.touchicon.plugin.php
```diff
@@ -43,33 +60,45 @@ public function SettingsController_TouchIcon_Create($Sender) {
...
             }
          } catch (Exception $ex) {
             $Sender->Form->AddError($ex->getMessage());
          }
-         
+
          $Sender->InformMessage(T('TouchIconSaved', "Your icon has been saved."));
```
tburry: Just use "Your icon has been saved." as the translation code.

PR #208, opened 2015-04-22
TouchIcon revisions

linc:
Addendum to #207

### DIFF
--- a/plugins/TouchIcon/class.touchicon.plugin.php
+++ b/plugins/TouchIcon/class.touchicon.plugin.php
@@ -45,7 +45,7 @@ class TouchIconPlugin extends Gdn_Plugin {
    public function Structure() {
       // Backwards compatibility with v1.
       if (C('Plugins.TouchIcon.Uploaded')) {
-         SaveToConfig('Garden.TouchIcon', PATH_ROOT.'/uploads/TouchIcon/apple-touch-icon.png');
+         SaveToConfig('Garden.TouchIcon', 'TouchIcon/apple-touch-icon.png');
          RemoveFromConfig('Plugins.TouchIcon.Uploaded');
       }
    }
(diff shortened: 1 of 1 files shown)


plugins/TouchIcon/class.touchicon.plugin.php
```diff
@@ -95,7 +96,7 @@ public function SettingsController_TouchIcon_Create($Sender) {
     * @return string Path to icon
     */
    public function getIconPath() {
-      $Icon = C('Garden.TouchIcon') ? Gdn_Upload::Url(C('Garden.TouchIcon')) : self::DEFAULT_PATH;
+      $Icon = C('Garden.TouchIcon') ? Gdn_Upload::Url(C('Garden.TouchIcon')) : Url(self::DEFAULT_PATH);
```
tburry: This should be `Asset()`, not `Url()`.

plugins/TouchIcon/class.touchicon.plugin.php
```diff
@@ -95,7 +96,7 @@ public function SettingsController_TouchIcon_Create($Sender) {
     * @return string Path to icon
     */
    public function getIconPath() {
```
tburry: If this returns a URL and not a path then maybe the method should be renamed.

plugins/TouchIcon/class.touchicon.plugin.php
```diff
@@ -45,7 +45,7 @@ public function OnDisable() {
    public function Structure() {
       // Backwards compatibility with v1.
       if (C('Plugins.TouchIcon.Uploaded')) {
-         SaveToConfig('Garden.TouchIcon', PATH_ROOT.'/uploads/TouchIcon/apple-touch-icon.png');
+         SaveToConfig('Garden.TouchIcon', 'TouchIcon/apple-touch-icon.png');
```
tburry: This is wrong. It should be blank for the default path or else you will be calling `Gdn_Upload::Url()` on a path that isn't in the uploads folder.

linc: No, the 1.0 version always saved the user-uploaded icon as this name in the uploads folder. The default is a constant.

The `Plugins.TouchIcon.Uploaded` config is no longer used; we're using it here to detect preexisting installs.

linc:
Updated

PR #209, opened 2015-05-03
Adds source to feed

adrianspeyer:
This makes sure the original feed gets credit with a link back.


linc:
Let's put this in its own div instead of using a `<br>` so that it can be styled different or hidden if desired.

PR #211, opened 2015-05-04
SteamConnect: Don't set username on existing users

### LINKED ISSUE #210: SteamConnect always overwrites forum name
Name should only be used on registration.

initvector:
Users' Vanilla usernames were being reset to their current username on Steam every time they authenticated with Steam.  This fix checks to see if the user already has an established Steam ID connection and avoids resetting their Vanilla username if found.

Fixes #210


linc:
You are my new favorite.

PR #212, opened 2015-05-04
Enable spoilers in PMs the rest of the way

linc:
NBBC + Spoilers + BBCode inputformatter required.


beckyvb:
Spoilers are still being revealed in the mebox messages dropdown. This is probably for another pull request.

beckyvb:
This one is also good to go, I believe.

linc:
Personally I think putting a spoiler in a PM is the dumbest idea ever. :D

PR #213, opened 2015-05-19
Akismet: Adding new service parameters

initvector:
Added comment_type, blog_lang and blog_charset.

Should not be merged until May 22nd at the earliest.

### DIFF
--- a/plugins/Akismet/class.akismet.php
+++ b/plugins/Akismet/class.akismet.php
@@ -207,6 +207,7 @@ class Akismet
 
    /**
     * Specify the character encoding for the form values included in comment_* parameters.
+    * 
     * @param string $blogCharset Character set for parameters.
     */
    public function setBlogCharset($blogCharset) {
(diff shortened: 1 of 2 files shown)


tburry:
Can you check the history of this plugin? We made a change a while back that was similar and it caused all spam to make it through detection and was a complete nightmare.

initvector:
52ff341ae5df5e9b25bf65ff63422aa43a6fade5 added comment_type, but set it to "forum", instead of "forum-post" as recommended [here](http://blog.akismet.com/2012/06/19/pro-tip-tell-us-your-comment_type/).

This PR sets comment_type to forum-post and only adds it for Discussion and Comment checks (the previous way tacked it onto every request).  This is in addition to adding language and charset parameters.

tburry:
The way to role this out would be to:

1. Add a config setting, but make the plugin behave as it currently does.
2. Find a friendly forum like penny arcade and flip that setting for a day.
3. See what happens.

Trust me. We don't want spam to start flooding in by accident.

linc:
Agreed. I have this running on my sites but that is mostly limited to vetting registrations, not posts, since my sites are approval registration. Have support identify a couple problem sites that would happily try an alternative.

PR #214, opened 2015-05-25
Rename the en-CA locales to en.

tburry:
Although en-CA will be canonicalized as en, this may not always be the case.


linc:
RANDO FIX IN YOUR PR ALERT :rotating_light:

PR #222, opened 2015-06-12
Replace spoiler body with text in messages popin and /messages/inbox.

beckyvb:
Fixes issue where spoiler body shows up in messages popin and messages/inbox excerpts.
Relies on events from https://github.com/vanilla/vanilla/pull/2829


tburry:
I think I see you doing a separate preg_replace for each format. You need to do a regular replace when the excerpt is converted to html, but before it's converted to plain text.

PR #232, opened 2015-07-08
Don't remove Answers from the Comments array. 

beckyvb:
Don't remove accepted comments from the Comments array. 
Instead, don't render accepted comments.
Fixes bug where attached files are not rendering for 'Answers'.
Also allow file uploads in New Question form.
Relies on event in https://github.com/vanilla/vanilla/pull/2883


tburry:
We don't skip the rendering of the comments here which I think I said on another PR. The data is what gets modified, not the rendering. Think of the views as being as logicless as possible.

PR #235, opened 2015-07-12
Add permission check to DiscussionOptions

bleistivt:
Currently everyone can change a discussion to a question and vice versa


linc:
Thanks!

PR #239, opened 2015-07-20
Render attachment as anchor in IPBFormatter if ImageWidth not present

initvector:
Attachments were previously always displayed as images when using IPBFormatter.  This update takes the updated logic from NBBC and applies it to DoAttachment in IPBFormatter to only display as an image when ImageWidth is present as an attribute for the Media record.


linc:
Use the new coding standards; so `va()` instead of `GetValue()`, and lowercase `anchor()`.

PR #240, opened 2015-07-21
Spacened and Casened the Akismet plugin, did not update the code comments.

nervoustwit:


linc:
:+1: for meritorious casening combat but no bonus points awarded.

PR #242, opened 2015-07-21
The Casening and Spacening come to IPB Formatter

initvector:


linc:
Variables too, eh? Ballsy.

Your double-to-single quote conversion is noted with head scratching. :hamster:

PR #244, opened 2015-07-27
Spacened and casened CivilTonguePlugin

nervoustwit:


plugins/CivilTongueEx/class.civiltongueex.plugin.php
```diff
@@ -2,357 +2,357 @@
...
+    }
+
+    /**
+     * Clean up activities and activity comments.
+     *
+     * @param Controller $Sender
+     * @param array $Args
+     */
+    public function activityController_render_before($Sender, $Args) {
+        $User = getValue('User', $Sender);
```
linc: Switch these to just `val()`, the new-and-improved version of that function.

plugins/CivilTongueEx/class.civiltongueex.plugin.php
```diff
@@ -2,357 +2,357 @@
...
+                }
+            }
+        }
+        return $Patterns;
+    }
+
+
+    public function setup() {
+        // Set default configuration
+        SaveToConfig('Plugins.CivilTongue.Replacement', '****');
```
linc: Lowercase

linc:
Just my two minor things above, ready to merge after that.

PR #250, opened 2015-08-20
Enhance jsconnect’s add/edit functionality

tburry:
* Build up the add/edit form programmatically with Gdn_Form->simple().
* Throw an event so that plugins can add controls to the form.
* Add JsConnectPlugin->addControl() so that plugins can more easily add controls.
* Standardize the saving of jsConnect connections to use proper form saving.


plugins/jsconnect/class.jsconnect.plugin.php
```diff
@@ -24,6 +24,15 @@ class JsConnectPlugin extends Gdn_Plugin {
 
     /// METHODS ///
 
+    public function addControl($key, $item) {
```
linc: Hark, a new method with no docblock. I name thee Scoundrel and bite my thumb at Thou.

PR #252, opened 2015-08-30
MeAction: Remove all assets

bleistivt:
- This plugin works just fine without the JS and CSS file. Apparently
this was rolled into the core formatter at some point.
- Updated the remaining code to match new coding standards.

### DIFF
--- a/plugins/MeAction/class.meaction.plugin.php
+++ b/plugins/MeAction/class.meaction.plugin.php
@@ -1,40 +1,29 @@
-<?php if (!defined('APPLICATION')) exit();
+<?php
 
 $PluginInfo['MeAction'] = array(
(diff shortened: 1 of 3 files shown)


linc:
If all this plugin does is flip a config toggle, it can be deleted.

bleistivt:
There is no UI for this in the dashboard though.

linc:
Fair enough. Seems like a case where it should be auto-on and we let people dig for the config if they really hate it tho. On the other hand I know @tburry dislikes the /me action because it makes theming more challenging by making the post area nonstandard.

linc:
Probably wanna increment the minor version number when making changes like this to plugins too.

PR #255, opened 2015-08-30
jsConnect: Remove livequery call

bleistivt:
Remove the livequery call and just initialize the `.JsConnect-Container` on document.ready.

I couldn't find a place where the `.JsConnect-Container` div could be added dynamically (see `JsConnectPlugin::connectButton()`).
This call has been in the plugin since the beginning, so it was most likely just defensive programming.


linc:
I believe your assessment is correct here. I can't think of a scenario where that'd happen.

linc:
Todd elaborated for me:

> Livequery calls the appropriate callback for each matched element individually. Just calling it directly does them all as a group.

linc:
I've reverted this on master for now.

PR #266, opened 2015-10-12
Missing translation in answer accepted activity text

supercosh:


linc:
Typo: `HeadlingFormat`

I think it would be best to just skip the slug here entirely (we don't use them for new translations anymore before it makes them difficult to translate) and just wrap the existing string in `t()`.

Issue #268, opened 2015-10-20
NBBC breaks search results when enabled.


valeryan:
NBBC returns all results with html characters encoded. the `searchExcerpt()` function in Vanilla's core `functions.render.php` expects html characters to be actual html. When the excerpt is grabbed from a string that is parsed by NBBC it can split html tags because they are encoded and are not recognized as an html tag.  This causes the search page layout to break when broken or incomplete html tags are included in the search results. 

I think either NBBC should be updated to return results with `html_entity_decode()`, or the `searchExcerpt()` function should run `html_entity_decode()` on all inputs before breaking out the excerpt.

valeryan:
I looked into this a little deeper to try and lock down why the NBBC breaks search results. The problem is that list items that are added by NBBC are not getting stripped out. In the [plaintText()](https://github.com/vanilla/vanilla/blob/master/library/core/class.format.php#L1012) function you are stripping list items to prevent it from breaking the search page but this misses nbbc's list items because they are encoded. Extra list items really feeds a kitten to an ATM if you know what I mean. I think it might be a good idea to do something like the [replaceSpoilers()](https://github.com/vanilla/vanilla/blob/master/library/core/class.format.php#L983) for list items and robustify the function to also strip out encoded list items to target this issues specifically.

PR #269, opened 2015-10-20
Wrap Format() return in html_entity_decode()

valeryan:
The built in parser for bbcode in Vanilla core returns parsed code with HTML tags decoded. This will make the NBBC parser return decoded html tags in the parser's results. This will prevent the search system from occasionally returning broken search results with NBBC enabled.

This will fix the issue I described in #268

### DIFF
--- a/plugins/NBBC/class.nbbc.plugin.php
+++ b/plugins/NBBC/class.nbbc.plugin.php
@@ -219,7 +219,7 @@ EOT;
    public function Format($Result) {
       $Result = str_replace(array('[CODE]', '[/CODE]'), array('[code]', '[/code]'), $Result);
       $Result = $this->NBBC()->Parse($Result);
-      return $Result;
+      return html_entity_decode($Result);
    }
 
    protected $_Media = NULL;
--- a/plugins/NBBC/class.nbbc.plugin.php
+++ b/plugins/NBBC/class.nbbc.plugin.php
@@ -23,7 +23,7 @@ $PluginInfo['NBBC'] = array(
 Gdn::FactoryInstall('BBCodeFormatter', 'NBBCPlugin', __FILE__, Gdn::FactorySingleton);
 
 class NBBCPlugin extends Gdn_Plugin {
-   
+
    public $Class = 'BBCode';
 
    /// CONSTRUCTOR ///
@@ -34,17 +34,17 @@ class NBBCPlugin extends Gdn_Plugin {
 
    /// PROPERTIES ///
    /// METHODS ///
-   
+
    public function DoAttachment($bbcode, $action, $name, $default, $params, $content) {
       $Medias = $this->Media();
       $MediaID = $content;
       if (isset($Medias[$MediaID])) {
          $Media = $Medias[$MediaID];
 //         decho($Media, 'Media');
-         
+
          $Src = htmlspecialchars(Gdn_Upload::Url(GetValue('Path', $Media)));
          $Name = htmlspecialchars(GetValue('Name', $Media));
-         
+
          if (GetValue('ImageWidth', $Media)) {
             return <<<EOT
 <div class="Attachment Image"><img src="$Src" alt="$Name" /></div>
@@ -53,7 +53,7 @@ EOT;
             return Anchor($Name, $Src, 'Attachment File');
          }
       }
-      
+
       return Anchor(T('Attachment not found.'), '#', 'Attachment NotFound');
    }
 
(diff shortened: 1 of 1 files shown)


linc:
The security implications of this need to be considered by the reviewer.

linc:
Core's bbcode parser passes the output thru `Gdn_Format::htmlFilter()` to make sure it's safe. My understanding of the situation currently is that:

* Without NBBC, you are allowed to use HTML but it is filtered. 
* With NBBC, all HTML becomes escaped.

To simply change this without adding a clause to run the `htmlFilter()` on the results when using a bbcode parser would open an XSS vector. That is a prerequisite for this change.

valeryan:
I looked at how results are returned in class.format.php and did the same call as the core parser. Maybe that meets your needs. 

however, I feel that it would actually be a better fix to refactor the  bbcode parser in Gdn_Format class to get the results form either the internal parser or a plugin parser and do your security filtering on all results so that a plugin doesn't have the ability to add this attack vector. 

I think it might work something like this in pseudo code:

```
function bbCode($mixed) {
// pseudo code for inside else starting line 262
$formatted = false;
if ($bbcodeformatter) {
    $result = bbcodeformatter->format($mixed)
    // other stuff happens
   $formatted = $results;
  // no return here
}
if (! $formatted) {
  // do normal stuff  in the $formatter logic sans the final return
  $formatted = $mixed2
}
// try catch handled possible errors to this point
// it should not hurt to run all results through decode
// and your security filter. This way the core controls the
// format of results instead of plugin. 
$result = html_entity_decode($formatted);
return Gdn_Format::html($result);
}
```
Anyway thats my thoughts on the matter. Let me know what you think...

linc:
I'd actually like to refactor NBBC into a vendor package within Vanilla and simply use that all the time. I don't think this PR is where we'll attempt that, however.

linc:
@initvector This looks good to me but I'd appreciate a consult.

plugins/NBBC/class.nbbc.plugin.php
```diff
@@ -219,7 +219,8 @@ function DoURL($bbcode, $action, $name, $default, $params, $content) {
    public function Format($Result) {
       $Result = str_replace(array('[CODE]', '[/CODE]'), array('[code]', '[/code]'), $Result);
       $Result = $this->NBBC()->Parse($Result);
-      return $Result;
```
valeryan: I apologies for the extra changes. I did not realize that my psr-2 fixer plugin had kicked in on this file. It makes it a little difficult to zero in on the changed code. I added this comment to highlight the change for you guys.

plugins/NBBC/class.nbbc.plugin.php
```diff
@@ -219,7 +219,8 @@ function DoURL($bbcode, $action, $name, $default, $params, $content) {
...
+      $Mixed = html_entity_decode($Result);
+      return Gdn_Format::html($Mixed);
```
valeryan: I apologies for the extra changes. I did not realize that my psr-2 fixer plugin had kicked in on this file. It makes it a little difficult to zero in on the changed code. I added this comment to highlight the change for you guys.

linc: No biggie, if you add ?w=1 to a GitHub diff it kills all the whitespace. I use it with abandon. :)

initvector:
My primary concern here is users will no longer be able to use HTML-encoded entities in their code blocks when using BBCode.  Currently, `[code]&lt;Testing this &amp; that.&gt;[/code]` renders as `&lt;Testing this &amp; that.&gt;`, using core BBCode parsing and NBBC.  With this update, that same code renders as `<Testing this & that.>`.  The contents of the code block aren't protected any more, since `html_entity_decode` is run on the full result of NBBC's parsing.

To reiterate the issue: When NBBC is parsing BBCode with its current configuration, it slaps `htmlspecialchars` on just about everything.  When we have a post with a mixed format of BBCode with HTML elements, the special characters (<, >, ', ", &) are encoded and what markup was there is no longer renderable by the browser.

valeryan:
@initvector @linc the issue #268 is all about search excerpts becoming broken. In the issues I suggested that I saw two ways to handle this. The first was to handle the issue in this addon. The other would be to make the `searchExcerpt()` in `functions.render.php` always convert back to html entities before trying to split it out so that the html will be recognized and the split will not happen in the middle of encoded tags. I think this may be the better options because it would make sure that no other addon could break the search results.

plugins/NBBC/class.nbbc.plugin.php
tburry: The call to **Gdn_Format::html()** will strip some css classes out that are appropriate for the BBCode parser.

tburry:
I think the searchExcerpt() should be robustified like @valeryan suggests.

valeryan:
I updated the issues this references with my latest findings. I am going to self close this PR as I don't think it really gets to the heart of the problem.

PR #275, opened 2015-11-16
Add missing edit button on QandA.

nervoustwit:


nervoustwit:
When a user clicks on Preview while posting a question there was no way for the user to go back to editing the question as you can do on discussions or comments.

PR #276, opened 2015-11-16
Disqus SSO joins open source

linc:


bleistivt:
All social sign in plugins re-implement similar curl logic.
I think a curl wrapper function should be added to the core instead.

And the Disqus logo is outdated:
https://disqus.com/brand/

linc:
Both good points! But I don't think either is cause to delay its release.

PR #280, opened 2015-11-25
Check of conversation model is available before instantiating in ignore plugin.

nervoustwit:


nervoustwit:
When you run a forum without the Conversation application off it breaks the ignore plugin because it tries to erase all conversations with the ignored user.

PR #296, opened 2016-02-18
Pockets: enable repeat on discussion table view

linc:
Currently, using the "Repeat Every" or "Given Indexes" option with Location "Between Discussions" does not work when `Vanilla.Layout.Discussions` is set to `table`. This is because the wrapper is set to `<li>` for those elements by default. This pull request makes Pockets aware of `Vanilla.Layout.Discussions` and uses a full-width table row instead.

### DIFF
--- a/plugins/Pockets/class.pockets.plugin.php
+++ b/plugins/Pockets/class.pockets.plugin.php
@@ -59,7 +59,10 @@ class PocketsPlugin extends Gdn_Plugin {
 
         // Switch our HTML wrapper when we're in a table view.
         if (c('Vanilla.Discussions.Layout') == 'table') {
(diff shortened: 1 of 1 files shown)


linc:
Also requires: https://github.com/vanilla/vanilla/pull/3547

initvector:
The table row is full-width, but the column the pocket goes into isn't on just about every browser except Firefox, seemingly the only browser that supports `colspan="0"`.

A discussion row might have five columns: Name, Started By, Replies, Views and Most Recent.  As it stands, the current code would drop a table column into the row that is only as wide as the first column: Name.

Our CSS sets `table-layout: fixed`, so we can't set the column's `colspan` attribute to some large number.

I think our only option is to attempt to determine the number of columns that might be present in the discussions table and use that for an absolute `colspan` value.

PR #297, opened 2016-02-18
Feature/QnA casening

DaazKu:
Fix function naming, PHP constants and indentation.

PR #298, opened 2016-02-19
Add optional points feature to Q&A

DaazKu:
Depends on https://github.com/vanilla/addons/pull/297

Adds configuration options to the QnA plugin.

* Award user points -> On/Off
* Points per answers (only for users' first answer to that question) -> uint
* Points per accepted answers -> uint


DaazKu:
-I kept the ``c($name, false);`` functions because even if not supplying the false parameters will have the same effect I find it clearer that way.
-I also kept the touchConfigs in setup because every call to ``c()`` have a default value supplied. Also, because the new feature is not active by default, the values will end up being written to the config file anyway (upon activation).

DaazKu:
Closing this.
Cleaner version of this PR -> #302

PR #302, opened 2016-02-26
Add points award feature to the QnA plugin.

DaazKu:
Adds configuration options to the QnA plugin.

Award user points -> On/Off
Points per answers (only for users' first answer to that question) -> uint
Points per accepted answers -> uint


linc:
Please don't mix variable casening with a PR this big in the future.

linc:
OK last thing: bump the version up to 1.3 since we're adding a feature

PR #304, opened 2016-03-03
New plugin PrefixDiscussionFilter

DaazKu:
Allow filtering by prefixes.

* Add 2 default filters
 * Has prefix
 * No prefix
* Generate a list of filters using existing prefixes.

### DIFF
--- a/plugins/PrefixDiscussionFilter/class.prefixdiscussionfilter.plugin.php
+++ b/plugins/PrefixDiscussionFilter/class.prefixdiscussionfilter.plugin.php
@@ -75,8 +75,8 @@ class PrefixDiscussionFilterPlugin extends Gdn_Plugin {
      */
     public function discussionModel_discussionFilters_handler($sender, $args) {
         DiscussionModel::addFilterSet('prefix', 'Prefixes');
-        DiscussionModel::addFilter('has-prefix', 'Has prefix', ['d.Prefix IS NOT NULL' => null], 'test', 'prefix');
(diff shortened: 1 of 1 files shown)


DaazKu:
Depends on https://github.com/R-J/PrefixDiscussion/pull/2

plugins/PrefixDiscussionFilter/class.prefixdiscussionfilter.plugin.php
```diff
@@ -0,0 +1,113 @@
...
+        $usedPrefixesResult =
+            Gdn::sql()
+                ->select('Prefix')
+                ->from('Discussion')
+                ->where('Prefix IS NOT NULL')
+                ->get()
+                ->resultArray();
+
+        foreach($usedPrefixesResult as $row) {
+            if (!isset($currentPrefixes[$row['Prefix']])) {
```
initvector: For the sake of legibility, set a new variable to the value of `$row['Prefix']` and use that.

plugins/PrefixDiscussionFilter/class.prefixdiscussionfilter.plugin.php
```diff
@@ -0,0 +1,113 @@
...
+                $currentPrefixes[$row['Prefix']] = $row['Prefix'];
+            }
+        }
+
+        natsort($currentPrefixes);
+        foreach($currentPrefixes as $prefix) {
+            DiscussionModel::addFilter('prefix-'.$this->stringToSlug($prefix), $prefix, ['d.Prefix' => $prefix], 'test2', 'prefix');
```
initvector: We need a better value for `$group` instead of "test2".

PR #307, opened 2016-03-09
Add option for redirect when changing locales in Multilingual.

nervoustwit:
Add a third parameter to the setlocale function so that a flag can be tacked on to the end of the redirect on success. Can be useful for triggering javascript and/or messaging.

### DIFF
--- a/plugins/Multilingual/class.multilingual.plugin.php
+++ b/plugins/Multilingual/class.multilingual.plugin.php
@@ -139,7 +139,7 @@ class MultilingualPlugin extends Gdn_Plugin {
     /**
      * Allow user to set their preferred locale via link-click.
      */
-    public function profileController_setLocale_create($Sender, $locale, $TK) {
+    public function profileController_setLocale_create($Sender, $locale, $TK, $appendRedirect = null) {
         if (!Gdn::Session()->UserID) {
             throw PermissionException('Garden.SignIn.Allow');
         }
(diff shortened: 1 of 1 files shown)


linc:
This is not safe. It's introducing an open redirect. Any data that involves redirecting the user needs to be validated and/or whitelisted.

plugins/Multilingual/class.multilingual.plugin.php
```diff
@@ -157,8 +157,15 @@ public function profileController_setLocale_create($Sender, $locale, $TK) {
...
+        // Create a whitelist of the configured locales or other acceptable variables that you could want to pass;
+        $redirectVariableWhitelist = array_merge(c('EnabledLocale'), array('changed', true, 1));
+
+        $successRedirect = $_SERVER['HTTP_REFERER'];
+        if ($appendRedirect && in_array($appendRedirect, $redirectVariableWhitelist)) {
+           $successRedirect = (stripos($_SERVER['HTTP_REFERER'], '?') === false) ? $_SERVER['HTTP_REFERER'] . "?localechange=" . $appendRedirect : $_SERVER['HTTP_REFERER'] . "&localechange=" . $appendRedirect;
+        }
         // Back from whence we came.
-        Redirect($_SERVER['HTTP_REFERER']);
+        Redirect($successRedirect);
```
linc: Switch this to `safeRedirect()`.

linc: There is a convention in Vanilla of simply using a `Target` parameter to specify a redirect. Will that not work here?

That $TK in the params should get removed in the future, so adding more URL params after it is problematic.

PR #314, opened 2016-03-29
Switch over to https for youtube links.

DaazKu:
Using http prevents the video from being rendered on https sites.
Also, youtube do not let you use http anymore so I didn't bother using the current site protocol.


initvector:
We've folded NBBC into core.  Is this something that needs to be fixed there too/instead?

DaazKu:
Yes. I will do the proper fix there.

PR #315, opened 2016-03-29
Points per accepted answer give the wrong value.

DaazKu:
Fix the number of points awarded from the QnA plugin when the Reaction plugin is also enabled.

### DIFF
--- a/plugins/QnA/class.qna.plugin.php
+++ b/plugins/QnA/class.qna.plugin.php
@@ -636,6 +636,11 @@ class QnAPlugin extends Gdn_Plugin {
 
                     case 'Accepted':
                         $Change = 1;
+
+                        if (!$this->Reactions && c('QnA.Points.Enabled', false) && $Discussion['InsertUserID'] != $Comment['InsertUserID']) {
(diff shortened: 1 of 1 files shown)


plugins/QnA/class.qna.plugin.php
```diff
@@ -630,7 +637,7 @@ public function discussionController_QnA_create($sender, $args) {
                     case 'Accepted':
                         $Change = 1;
 
-                        if (c('QnA.Points.Enabled', false) && $Discussion['InsertUserID'] != $Comment['InsertUserID']) {
+                        if (!$this->Reactions && c('QnA.Points.Enabled', false) && $Discussion['InsertUserID'] != $Comment['InsertUserID']) {
```
linc: Won't this property be entirely missing and generate a notice if Reactions is off?

DaazKu: You mean `$this->Reactions` ????

linc: Yup - missed that. Got my inheritance confused.

PR #316, opened 2016-03-30
Add event to correct for question count based on sub communities.

nervoustwit:
Create a hook for the optional adjusting of the number of questions displayed in the QnA plugin based on, for example, the current sub community in the cloud hosted environment.

This will fix the issue https://github.com/vanilla/multisite/issues/28

A handler was added to the subcommunities plugin to handle this event in PR https://github.com/vanilla/multisite/pull/29/

### DIFF
--- a/plugins/QnA/class.qna.plugin.php
+++ b/plugins/QnA/class.qna.plugin.php
@@ -973,13 +973,23 @@ class QnAPlugin extends Gdn_Plugin {
      * @return int
      */
     public function getUnansweredCount() {
-        $questionCount = Gdn::cache()->get('QnA-UnansweredCount');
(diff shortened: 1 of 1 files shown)


plugins/QnA/class.qna.plugin.php
```diff
@@ -973,13 +973,15 @@ public function discussionsController_beforeBuildPager_handler($sender, $args) {
      * @return int
      */
     public function getUnansweredCount() {
-        $Count = Gdn::cache()->get('QnA-UnansweredCount');
-        if ($Count === Gdn_Cache::CACHEOP_FAILURE) {
+        $questionCount = Gdn::cache()->get('QnA-UnansweredCount');
```
tburry: This won't work because the cache key is global.

PR #355, opened 2016-07-25
Add redirections for Ning

DaazKu:
More redirections!


initvector:
I think we're gonna a different approach here.

PR #364, opened 2016-08-19
Take Garden.Registration.ConfirmEmail into account in the WelcomePost plugin

DaazKu:
Redirect to the welcomepost page after email confirmation if enabled.


JulieVanilla:
Does it work for Approval style registration? I imagine Invitation and Connect styles are fine, but unsure on Approval, since it requires action from the admin before they generally have access to post.

DaazKu:
I'm pretty sure it will not work with approval.

linc:
@DaazKu Add a note to the plugin description and/or README to clarify that either way please.

PR #366, opened 2016-08-25
Add t() to Unanswered Questions title

### LINKED ISSUE #103: Title Page  of view Unanswered Questions does not seem to be translateable
Possible change , unless you have a better one that we are missing.  

   Gdn::Controller()->Title('Unanswered Questions');

to

Gdn::Controller()->Title(T('Unanswered Questions'));

see:  http://vanillaforums.org/discussion/comment/209924/#Comment_209924

linc:
Closes #103


linc:
This translation is already in locales. The net effect is going to be the page title changing to simply 'Unanswered' which I think is fine.

PR #371, opened 2016-08-29
Enhance Akismet to validate your key

### LINKED ISSUE #149: Askimet: Bad API key prevents new discussion creation
{"Code":0,"Exception":"The Wordpress API key passed to the Akismet constructor is invalid. Please obtain a valid one from http:\/\/wordpress.com\/api-keys\/","Class":"Exception"}
apierror

linc:
Closes #149 

* An invalid key blocks all unverified users from posting, which is a catastrophe.
* Akismet already provides the mechanism for verifying a key, so let's use it.
* Refactor the Akismet builder to allow alternate key testing.

### DIFF
--- a/plugins/Akismet/class.akismet.plugin.php
+++ b/plugins/Akismet/class.akismet.plugin.php
@@ -29,18 +29,27 @@ class AkismetPlugin extends Gdn_Plugin {
     public static function akismet() {
         static $Akismet;
         if (!$Akismet) {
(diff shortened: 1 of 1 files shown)


linc:
Note that https://github.com/vanilla/vanilla/pull/4377 is needed for this to BLOCK an invalid key, but this will at least notify the user they messed up without it.

PR #375, opened 2016-09-01
Fix Q&A interactions with other plugins

DaazKu:
- Move the `questionCount` hook so that it works when caching is enabled.
- Show unanswered questions from groups. 
  - *The group category has the `HideAllDiscussions` flag set to true. This means that the category won't be returned by `CategoryModel::categoryWatch();`. To circumvent that we explicitly set all categories on the discussion controller. That way we will get the unanswered questions from everywhere.*
- Add a hook to allow the categories from previous point to be modified. Needed by Subcommunities for example.


linc:
`CategoryWatch` also filters on permissions. Does this no longer take permissions into account, then?

DaazKu:
Still effective.

https://github.com/vanilla/vanilla/blob/master/applications/vanilla/controllers/class.discussionscontroller.php#L132

linc:
Sweet

PR #388, opened 2016-09-12
Update coding standards in jsConnect

linc:
No functional changes.

Most interesting thing I did was removed a logic block setting `$RegisterLink` where it was never used.


nervoustwit:
One thing:

- plugins/jsconnect/views/jsconnect.php
- plugins/jsconnect/functions.jsconnect.php 

don't have line breaks at the end of file.

Issue #408, opened 2016-09-28
CivilTongueEx filtering is broken for LastDiscussions.


DaazKu:
The plugin filters "[Categories](https://github.com/vanilla/addons/blob/3b3e39e2554e97e3f690e32b2c44d3b8935ce195/plugins/CivilTongueEx/class.civiltongueex.plugin.php#L172)" but now it is named "CategoryTree"

[We also flatten that tree later on](
https://github.com/vanilla/vanilla/blob/5cdeb2c30776da6367f48c674539bbcf7fc8e236/applications/vanilla/views/categories/all.php#L18). This add children categories to the list which bypass the filter done in `_render_before`.

The plugin also needs an update for groups' events.

PR #419, opened 2016-10-18
Update CivilTongueEx to work on Groups and Events + bugfix on categories

DaazKu:
Depends on https://github.com/vanilla/internal/pull/737
Fixes https://github.com/vanilla/addons/issues/408

Fix filtering on the categories view.
Filter out groups and events fields.

P.S. Yeah I know it is a 2 in one.. don't crucify me.

### DIFF
--- a/plugins/CivilTongueEx/class.civiltongueex.plugin.php
+++ b/plugins/CivilTongueEx/class.civiltongueex.plugin.php
@@ -491,6 +491,8 @@ class CivilTonguePlugin extends Gdn_Plugin {
     /**
      * Replace bad words in the group list
      *
+     * Vanilla's proprietary group plugin hook.
+     *
      * @param SettingsController $sender Sending controller instance
      * @param array $args Event's arguments
      */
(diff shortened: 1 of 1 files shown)


linc:
We need to be careful when crossing the streams between open source and our proprietary code base. Not because the hooks are top secret (far from it), but because by not publishing Groups we have not "claimed" that namespace in the public arena. This could lead to very unintended consequences for an open source user that enables this plugin with a third-party addon that does claim it. See: https://vanillaforums.org/addon/groups-plugin and https://vanillaforums.org/get/eventcalendar-plugin-0.2.1

Fortunately, it doesn't appear that the Groups addon does anything substabtive enough to use its own controllers and events, and the Event Calendar one uses the entire title as a slug. Still, this is cutting things close. Keep these interactions in mind in the future. For this one, you might at least add a description line in the docblocks to note they hook into the cloud addons, not any open source ones. It's icky code pollution, but I don't have a better solution here presently.

PR #421, opened 2016-10-19
Author Selector: fix Javascript conflict that hides Editor bar in 2.3; update standards

linc:
- Update how Javascript initiates so it doesn't conflict with other addons.
- Spacening, casening, and docs.
- Version to 1.2.

### DIFF
--- a/plugins/AuthorSelector/js/authorselector.js
+++ b/plugins/AuthorSelector/js/authorselector.js
@@ -1,31 +1,27 @@
-var authorSelector = {
-    start: function($author) {
-        /// Author tag token input.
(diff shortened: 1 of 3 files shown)


plugins/AuthorSelector/js/authorselector.js
```diff
@@ -25,8 +22,10 @@ jQuery(document).ready(function($) {
             animateDropdown: false,
             tokenLimit: 1
         });
-    };
+    }
+}
```
DaazKu: Missing ;

linc:
Like that?

PR #434, opened 2016-11-17
Add warning to jsConnect dashboard when providers in test mode

### LINKED ISSUE #385: jsConnect should warn when test mode is enabled
If there is an active connection method for jsConnect that has test mode enabled, a big warning should appear at the top of the main jsConnect settings page.

initvector:
This update adds a warning message to the jsConnect area of the dashboard when a provider is in test mode.

Closes #385

### DIFF
--- a/plugins/jsconnect/views/settings.php
+++ b/plugins/jsconnect/views/settings.php
@@ -19,15 +19,10 @@ foreach ($this->Data('Providers') as $Provider) {
 ?>
 
 <?php if (count($inTestMode) > 0): ?>
-<div class="alert alert-warning padded"><?php echo t('Providers in test mode.', 'The following providers are in test mode:'); ?>
+<div class="alert alert-warning padded"><?php echo t('Providers in test mode.', 'The following providers are in test mode and are not secure.  Incoming connections will be accepted without verifying the source.'); ?>
     <ul>
     <?php foreach ($inTestMode as $testProvider): ?>
-        <li><?php echo anchor(
-            $testProvider['Name'],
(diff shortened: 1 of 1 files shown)


plugins/jsconnect/views/settings.php
```diff
@@ -9,7 +9,30 @@
...
+$inTestMode = [];
+foreach ($this->Data('Providers') as $Provider) {
+    if ($Provider['TestMode']) {
+        $inTestMode[] = $Provider;
+    }
+}
 ?>
+
+<?php if (count($inTestMode) > 0): ?>
+<div class="alert alert-warning padded"><?echo t('Providers in test mode.', 'The following providers are in test mode:'); ?>
```
beckyvb: missing a `php` here

initvector: Fixed!

beckyvb:
I'm not real wild about the anchors bringing up a popup to edit a connection. It doesn't seem intuitive that that's what the anchor will do (even with the title set). Also, the alert message could be a little more illuminating. What are the dangers of being test mode? Also, is `alert-danger` more appropriate here?

initvector:
> I'm not real wild about the anchors bringing up a popup to edit a connection. It doesn't seem intuitive that that's what the anchor will do (even with the title set).

Anchors have been axed.

> The alert message could be a little more illuminating. What are the dangers of being test mode?

Good point.  The "test mode" notice has been updated.

> Also, is alert-danger more appropriate here?

`alert-warning` feels right to me.  Everything is still functional, but the user will need to be careful.

PR #437, opened 2016-11-21
Addon installation section in README

DaazKu:
Let's add this before we get issues about it :)


linc:
This isn't accurate; It's not present in 2.3.

PR #438, opened 2016-11-22
[Troll] Fix the toggle option not being avalaible in the profile when there are no trolls

Caylus:
If there were no trolls, the option to mark an user as one would never appear.


DaazKu:
Could you update the DOC block to reflect that this function put a "toggle" and not a "remove" button?

Caylus:
Like this?

DaazKu:
Yes thank you

PR #440, opened 2016-12-01
QnA: Move structure-related methods to separate file and fix docblocks

linc:


plugins/QnA/class.qna.plugin.php
```diff
@@ -53,187 +59,25 @@ public function setup() {
...
-            }
-
-            // AcceptAnswer
-            $Rm->defineReactionType(array('UrlCode' => 'AcceptAnswer', 'Name' => 'Accept Answer', 'Sort' => 0, 'Class' => 'Positive', 'IncrementColumn' => 'Score', 'IncrementValue' => 5, 'Points' => $points, 'Permission' => 'Garden.Curation.Manage', 'Hidden' => 1,
-                'Description' => "When someone correctly answers a question, they are rewarded with this reaction."));
-        }
-
-        Gdn::structure()->reset();
+    public function structure() {
+        include dirname(__FILE__).'/structure.php';
```
DaazKu: __DIR__ would be shorted.
Just an FYI.

linc: Yeah I copy-pasted and that was the way to maintain older PHP compat.

linc:
Found a few other things to cleanup as well.

I really did not like the idea of a settings page save calling into `structure()` at all. This plugin is maximum weirdness with its interactions.

Issue #428, opened 2016-11-01
Answers authors points not recalculated properly when the answer is edited by an admin


beckyvb:
EDIT
---
The Q&A menu options did not recalculate points properly when the status was changed.

-----

QnA's handling of reactions assumes that only one user will be managing the QnA status. It relies on the toggling nature of Reactions. (i.e., reacting once enables, reacting twice disables). This goes through the roof when another user comes in and changes the status of a question. When two users edit the status of an answer, it results in the comment author getting double the points.

Steps to reproduce:
1. Ensure Reactions is enabled and QnA points are enabled.
2. Accept a comment as an answer.
3. With another (admin) user, edit the QnA status to 'Rejected' or 'Don't Know'

Result:
There are two Accepted Answer reactions on the comment, resulting in the author being awarded double the points.

Expected result:
The comment author loses their points for the Accepted Answer.

DaazKu:
The feature is utterly broken.

Points are not given on first answer.
Reactions point are not set properly when enabled/disabling the plugin.
Points are not given/removed properly when using the QnAOptions (admin tool)
Points are not given/removed properly when not using reactions.

PR #443, opened 2016-12-21
Fix Q&A point-awarding interactions with Reactions

DaazKu:
Fix https://github.com/vanilla/addons/issues/428

- [x] Change points properly using the QnAOptions (admin tool)
  - [x] With Reaction turned ON
  - [x] With Reaction turned OFF
- [x] Update Accepted Answer reaction points properly
  - [x] First setup with the Reaction plugin ON
  - [x] Subsequent setup with the Reaction plugin ON (Turn the QnA plugin on and configure the point feature then turn the plugin off, turn reaction ON, turn the QnA plugin back ON)

Check the UserPoint table.
If Reaction is OFF the only source should be QnA.
If Reaction is ON there should be both QnA (for first answer on a question) and Reaction (for accepted answer).

Using the QnAOptions tool should remove/add points properly.

### DIFF
--- a/plugins/QnA/class.qna.plugin.php
+++ b/plugins/QnA/class.qna.plugin.php
@@ -526,8 +526,10 @@ class QnAPlugin extends Gdn_Plugin {
                     include_once(Gdn::controller()->fetchViewLocation('reaction_functions', '', 'plugins/Reactions'));
                     $Rm = new ReactionModel();
 
+                    // Assume that the reaction is done by the question's owner
+                    $questionOwner = $Discussion['InsertUserID'];
                     // If there's change, reactions will take care of it
-                    $Rm->react('Comment', $Comment['CommentID'], 'AcceptAnswer', null, true);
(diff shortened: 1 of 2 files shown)


beckyvb:
This doesn't fix the original issue described here: https://github.com/vanilla/addons/issues/428 The original issue stems from multiple users changing the status of an answer and when a question is rejected by a second user, then Reactions does not remove the AcceptedAnswer reaction (as expected) but rather adds another AcceptedAnswer reaction for a second user. From what I can tell, this is where that occurs: https://github.com/vanilla/addons/blob/master/plugins/QnA/class.qna.plugin.php#L523 If you follow the steps to reproduce in the original issue, you should be able to reproduce.

beckyvb:
A user gains points for accepting their own answer on a question they started using reactions, but does not if reactions is off. I'm not sure which behaviour is better, but it should be consistent.

DaazKu:
They should not get points. I'll fix that too.

PR #445, opened 2017-01-16
Change the functionality of the Welcome Post plugin

nervoustwit:
This PR makes changes to the Welcome Post plugin, which sends new users to /post/discussion prompting them to introduce themselves to the community:

Removes the messaging from a javascript modal and em

- beds it in the page
- Changes the title of the page
- Instead of Cancel it says Skip (since the user has been redirected there while en route to another page)
- Supports users who have just created accounts via SSO


linc:
You need to add 'Skip' to the locales repo now.

PR #452, opened 2017-01-26
Removes MediaModel from FileUpload plugin

DaazKu:
Fixes https://github.com/vanilla/vanilla/issues/4055
Depends on https://github.com/vanilla/vanilla/pull/5058
**Run composer update after dependency has been merged!**

Removes media model and move it into core.
Moves functions, that were wrongly put into the MediaModel, to the FileUpload plugin.


linc:
Plugin needs to be updated to require a higher Vanilla version.

I'd amend #5058 (and any future PRs that add big things to core) to increment the version.

PR #453, opened 2017-01-30
Link to root of forum when opting out of Welcome Post

nervoustwit:
The Welcome Post gives the user the option to not post a "Welcome" post by clicking cancel which would send you to the default controller. This could have some unwanted consequences. This PR sends the user to / and assumes routing will make sure the user is going to the default page.


nervoustwit:
I went with isSafeURL because it is usually a target with in the site.

PR #471, opened 2017-03-21
Refactor SiteMap Plugin's code

slafleche:
Refactored code to match current standards.


plugins/Sitemaps/class.sitemaps.plugin.php
```diff
@@ -1,4 +1,6 @@
-<?php if (!defined('APPLICATION')) exit();
+<?php if (!defined('APPLICATION')) {
+    exit();
+}
```
DaazKu: One line plz

PR #472, opened 2017-03-22
Sitemap Plugin - Added event to filter categories

slafleche:
This plugin didn't work properly with plugins that changed the url structure of Vanilla. I've added an event that allows to filter the categories sent to the plugin.

### DIFF
--- a/plugins/Sitemaps/class.sitemaps.plugin.php
+++ b/plugins/Sitemaps/class.sitemaps.plugin.php
@@ -124,8 +124,8 @@ class SitemapsPlugin extends Gdn_Plugin {
     }
 
     /**
-     * @param Gdn_Controller $Sender
-     * @param type $args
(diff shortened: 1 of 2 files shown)


plugins/Sitemaps/class.sitemaps.plugin.php
```diff
@@ -127,7 +127,7 @@ public function utilityController_robots_create($Sender) {
      * @param Gdn_Controller $Sender
      * @param type $Args
      */
-    public function utilityController_siteMapIndex_create($Sender) {
+    public function utilityController_siteMapIndex_create($Sender, $args = []) {
```
DaazKu: Do not default the args.
You also need to fix the docblock :P 
```php
@param UtilityController $Sender Sending controller instance
@param array $args Event's arguments
```

Issue #468, opened 2017-03-20
Disqus: Callback URL must now match exactly


linc:
Disqus tightened the security checks so the `redirect_url` has to match Disqus' Callback URL field exactly. Vanilla adds `?Target=<category>` to the `redirect_url` so the URL changes every time, causing mismatches.

We either need a workaround or to remove the Target redirect, which is annoying but not the end of the world.

Reference: https://disqus.com/home/discussion/channel-discussdisqus/api_invalid_parameter_redirect_uri_should_match_predefined_callback_uri/

DaazKu:
I'm gonna try to do it this sprint.

PR #475, opened 2017-03-27
Remove redirect_uri form Disqus

DaazKu:
Fixes https://github.com/vanilla/addons/issues/468

The callback URL must now match exactly what is configured and the Application.
Since the callback is a mandatory field it is now useless to provide it.


DaazKu:
Seems like the API call still need the parameter!

PR #477, opened 2017-03-31
Optimize discussion polling in Participated plug-in

### LINKED ISSUE #462: Participated: Rework using 'Participated' flag in core
Currently it's using a query that does not scale.

@kaecyra may have more info.

initvector:
This update attempts to optimize how a user's "participated" discussions are queried. The UserDiscussion table is queried based on the value of the `Participated` field and the Discussion table is joined in.

The methods of joining in additional user and category data seemed outdated, so I've updated them to match what's currently used in `DiscussionModel::getWhere`.

Closes #462


DaazKu:
By removing the join on comments you changed the behaviour of the plugin.

Before you only saw discussion that you actually commented on. Now you also see discussions that you created but did not comment on.

PR #482, opened 2017-04-17
Update coding standards for Bump

R-J:
Includes fix for links in forums with subfolders


linc:
Avoid vertically aligning things generally (here, in the doc blocks).

PR #488, opened 2017-05-02
Add license information to PluginInfo array

R-J:
The license was missing from most of the plugins PluginInfo array. That key is needed when the plugin should be upload to open.vanillaforums.com


linc:
This was a good change, but it had the extreme misfortune of coming in just as all those plugins had their info changed to JSON. I'm gonna close this out because I think it's easier to do over than fix.

PR #494, opened 2017-05-15
Delete TouchIcon plugin

slafleche:
After changes to core have been merged: https://github.com/vanilla/vanilla/pull/5577, delete plugin.


initvector:
It's probably not a good idea to outright delete the plug-in so soon after migrating its functionality into core. Maybe deprecate, first.

It'd be different if this plug-in was in the core repo. You could safely assume the code was always present. However, you are making the assumption that whoever is running this plug-in will also be on bleeding-edge master of core. It's possible a user could lose this functionality if they updated the addons repo, but were using an older revision/release of Vanilla.

initvector:
I'm gonna close this out, since I'm pretty sure our desired course of action is to deprecated, not delete.

PR #505, opened 2017-05-31
Allow error messages in signature plugin to be translated

nervoustwit:
Error messages in the signature plugin are untranslatable. This PR will fix that .


plugins/Signatures/class.signatures.plugin.php
```diff
@@ -255,16 +255,15 @@ public function checkSignatureLength($Values, &$Sender) {
...
-            } elseif (is_int($maxImages) && $numMatches > $maxImages) {
-                $Sender->Form->addError('@'.formatString('You are only allowed {maxImages,plural,%s image,%s images}.',
-                        ['maxImages' => $maxImages]));
-
+            if (C('Plugins.Signatures.MaxNumberImages') === 'None' && $numMatches > 0) {
+                $Sender->Form->AddError(t('Images not allowed.'));
+            } else if ($numMatches > $max) {
+                $Sender->Form->AddError('@'.FormatString(t('You are only allowed {maxImages,plural,%s image,%s images}.'),
+                        array('maxImages' => $max)));
             }
```
linc: You reversed a bunch of coding standards here.

plugins/Signatures/class.signatures.plugin.php
```diff
@@ -255,16 +255,15 @@ public function checkSignatureLength($Values, &$Sender) {
      */
     public function checkNumberOfImages($Values, &$Sender) {
         $maxImages = self::getMaximumNumberOfImages();
+        $max = c('Plugins.Signatures.MaxNumberImages');
```
linc: 1) Is there a config still named this? I thought Stephane just changed the namespace for all these last month. 2) `$max` isn't enough context for a variable name when 'max' is used 4 times in 2 lines.

nervoustwit:
Messy PR that never should have been made. The problem will be solved with https://github.com/vanilla/addons/pull/507 . Closing this.

Issue #467, opened 2017-03-08
CivilTongue does not filter notifications


linc:
You can sneak in naughty words!!!1

slafleche:
In PR: https://github.com/vanilla/addons/pull/506

PR #506, opened 2017-06-01
Civil Tongue: Add filtering to activity & notifications

slafleche:
Hooked into "beforeSave" to filter out activity messages in Civil Tongue plugin.

Closes https://github.com/vanilla/addons/issues/467

### DIFF
--- a/plugins/CivilTongueEx/class.civiltongueex.plugin.php
+++ b/plugins/CivilTongueEx/class.civiltongueex.plugin.php
@@ -384,7 +384,7 @@ class CivilTonguePlugin extends Gdn_Plugin {
      * @param $Sender
      * @param $Args
      */
-    public function activityModel_BeforeSave_Handler($Sender, &$Args) {
+    public function activityModel_beforeSave_handler($Sender, &$Args) {
         $activity = val('Activity', $Args);
         setValue('Story', $activity, $this->replace(val('Story', $activity)));
     }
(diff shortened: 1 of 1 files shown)


plugins/CivilTongueEx/class.civiltongueex.plugin.php
```diff
@@ -379,6 +379,18 @@ public function notificationsController_informNotifications_handler($Sender, &$A
     }
 
     /**
+     * Cleanup Activity messages.
+     *
+     * @param $Sender
+     * @param $Args
+     */
+    public function ActivityModel_BeforeSave_Handler($Sender, &$Args) {
```
linc: camelCase not CapitalCase.

slafleche:
Yes I did test 
screen shot 2017-06-02 at 08 17 51

linc:
Look at the method names above and below this one and try again.

linc:
So, this is working because the value at the key of 'Activity' in this case happens to be an object, which is passed by reference. In all other cases (including if 'Activity' happened to be formatted as an array) this would fail because you're just setting in a value that is never returned or set on $Args. And, despite using `val()` as an abstraction (which is so we don't care if this is object or array) this makes it care - it explicitly breaks anyway as an array.

The correct way to close this loop is to create a third line that sets the value of `Activity` back on the args. But I would do this the more common way, which is to set `$sender->EventArguments['Activity'] = $activity;` manually rather than passing the `$Args` param by reference. We don't pass params as a reference anymore and it's bad to rely on that.

PR #480, opened 2017-04-05
Signature plugin update

slafleche:
Updated code style
Made logic more robust to various configurations

PR #509, opened 2017-06-20
Fix length check in Signatures addon to use new config

initvector:
#480 introduced a slew of updates to the Signatures addon, including updating some config keys. The problem is the new config keys aren't always being used. For example, [`SignaturesPlugin::checkSignatureLength` checks if a limit is configured by calling `SignaturesPlugin::getMaximumTextLength`](https://github.com/vanilla/addons/blob/7504d0ba53a242678885ce8992d038115d32567f/plugins/Signatures/class.signatures.plugin.php#L239) and then performs comparisons based on the value of the Plugins.Signatures.MaxLength config. The problem is `SignaturesPlugin::getMaximumTextLength` [uses the newer Signatures.Text.MaxLength config](https://github.com/vanilla/addons/blob/7504d0ba53a242678885ce8992d038115d32567f/plugins/Signatures/class.signatures.plugin.php#L791). If these two config values conflict, you can potentially encounter erroneous validation behavior when saving a signature.

This update alters some coding standards in `SignaturesPlugin::checkSignatureLength` and updates it to consistently use the value returned by `SignaturesPlugin::getMaximumTextLength` and therefore the new config key.

Please backport to [release/2017-Q2-4](https://github.com/vanilla/addons/tree/release/2017-Q2-4)

### DIFF
--- a/plugins/Signatures/class.signatures.plugin.php
+++ b/plugins/Signatures/class.signatures.plugin.php
@@ -228,21 +228,33 @@ class SignaturesPlugin extends Gdn_Plugin {
     }
 
     /**
-     * Checks signature length against Plugins.Signatures.MaxLength
+     * Determine if signature exceeds configured character limit.
      *
-     * @param $Values Signature settings form values
-     * @param $Sender Controller
+     * @param array $fields Signature settings form values
+     * @param ProfileController $sender
      */
-    public function checkSignatureLength($Values, &$Sender) {
+    public function checkSignatureLength($fields, &$sender) {
+        if (!property_exists($sender, 'Form') || !($sender->Form instanceof Gdn_Form)) {
+            return;
+        }
 
-
-        if (!is_null(self::getMaximumTextLength())) {
-            $Sig = Gdn_Format::to($Values['Plugin.Signatures.Sig'], $Sender->Form->getFormValue('Format'));
-            $TextValue = html_entity_decode(trim(strip_tags($Sig)));
+        $maxLength = self::getMaximumTextLength();
+        if ($maxLength !== null && $maxLength > 0) {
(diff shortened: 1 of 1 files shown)


plugins/Signatures/class.signatures.plugin.php
```diff
@@ -228,21 +228,33 @@ public function crossCheckSignature($Values, &$Sender) {
...
+            return;
+        }
 
-        if (!is_null(self::getMaximumTextLength())) {
-            $Sig = Gdn_Format::to($Values['Plugin.Signatures.Sig'], $Sender->Form->getFormValue('Format'));
-            $TextValue = html_entity_decode(trim(strip_tags($Sig)));
+        $maxLength = self::getMaximumTextLength();
+        if ($maxLength !== null && $maxLength > 0) {
+            $maxLength = intval($maxLength);
+            $format = config('Plugin.Signatures.Format', Gdn_Format::defaultFormat());
```
DaazKu: #480 failed to properly migrate all configurations. The correct fix would be to address that.

Anything with `Plugin.Signatures.` should be changed to `Signatures.`.

initvector: I'd really like to limit the scope of this PR to a specific issue. There is a known issue with this particular check and it can be easily tested to confirm resolution with lessened potential of introducing adverse side effects.

That said, I *think* this is actually the only config key that's still in the "Plugin" namespace. There are other key/value pairs in the "Plugin" namespace, but they're in form fields and user preferences.

DaazKu: Ok but make sure to create an issue about #480 not having properly migrated all config values.

It migrated `Plugins.Signatures` but not `Plugin.Signatures`...

Issue #521, opened 2017-06-30
Missing clear float for signature


slafleche:
The signature plugin needs to clear floats. When we have, for example, a right aligned text, followed by the signature, they overlap.

slafleche:
In PR https://github.com/vanilla/addons/pull/522

PR #522, opened 2017-06-30
Clear floats for signatures plugin

slafleche:
Signatures didn't clear floats, creating issues when the last element was floating right (like a right align for example)

Fixes: https://github.com/vanilla/addons/issues/521


Charrondev:
We should probably be clearing the message body eh? That would be a more robust solution.

slafleche:
Good idea! I'm closing this PR:

PR #529, opened 2017-07-07
[Pockets] Fix undefined offset notice in Pockets

Charrondev:
A notice was being thrown by this callback function `PocketsPlugin::pocketStringCb()` because it was being called outside of `preg_replace_callback()`. This PR:

- deprecates `pocketStringCb` function which should never have been marked as public
- moves the function into a lambda
- removes the extra call the callback

This should make it into 2.5 so I've added a task to delete the deprecated function in 2.6 - #530

### DIFF
--- a/plugins/Pockets/class.pockets.plugin.php
+++ b/plugins/Pockets/class.pockets.plugin.php
@@ -630,7 +630,7 @@ class PocketsPlugin extends Gdn_Plugin {
                 }
             };
 
-            $result = preg_replace_callback('`{{(\w+)}}`', $callback, $result);
+            $result = preg_replace_callback('/{{(\w+)}}/', $callback, $result);
         }
 
         return $result;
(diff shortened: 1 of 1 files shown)


plugins/Pockets/class.pockets.plugin.php
```diff
@@ -620,22 +620,31 @@ public static function pocketString($Name, $Data = null) {
...
      *
-     *
-     * @param null $Match
+     * @deprecated 2.4
+     * @param array $Match
      * @param bool|false $SetArgs
      * @return string
      */
     public static function pocketStringCb($Match = null, $SetArgs = false) {
+        deprecated('PocketsPlugin::pocketStringCb()');
```
DaazKu: __METHOD__ will yield the same result here

plugins/Pockets/class.pockets.plugin.php
```diff
@@ -620,22 +620,31 @@ public static function pocketString($Name, $Data = null) {
 
         if (is_array($Data)) {
             $Data = array_change_key_case($Data);
+            $callback = function ($matches) use ($Data) {
+                $yey = strtolower($matches[1]);
```
DaazKu: Never copy paste code without reviewing it first ;)

DaazKu: Also `use($Data)` but `isset($data[$key])` ????

PR #546, opened 2017-07-28
Refactor function names to respect our coding standard

DaazKu:
I used this script: https://gist.github.com/DaazKu/fc20843e3f7bd2e0311194c0e70dada5
- It's using the tokenizer.
- It skips class instantiation by checking that there is no `new` before the token. (It also skips namespace searching for the new keyword)
- It's also handling:
   - `SQL()` to `sql()`
   - `Controller_Event_Handler()` to `controller_event_handler()`
   - `AnythingElse()` to `anythingElse()`
- It renames function in comments. (That part is regex based but that should do)

### DIFF
--- a/plugins/Akismet/class.akismet.php
+++ b/plugins/Akismet/class.akismet.php
@@ -31,7 +31,7 @@
  *
  *  <b>Usage:</b>
  *  <code>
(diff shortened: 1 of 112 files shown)


DaazKu:
Cross link: vanilla/vanilla#5803

PR #552, opened 2017-09-04
Optionally add a default body to the Welcome Post

nervoustwit:
Presently the Welcome Post addon only supplies a default title to the welcome post, this PR adds a translation code so that forum admins can add a translation string to pre-fill the welcome post body.


DaazKu:
It should probably be documented in a README.md

Issue #551, opened 2017-09-01
Q&A: Doesn't work with Reactions' per-category point tracking


linc:
Q&A now allows you to award points for accepted answers. It does not appear to work with Reactions' ability to award points separately for some categories. This is a basic plugin conflict that should get resolved.

linc:
https://vanillaforums.teamwork.com/#tasks/6585265

PR #559, opened 2017-10-23
Fix QnA addon not tracking points per-category

DaazKu:
Fixes https://github.com/vanilla/addons/issues/551

There is a feature that you can enable that track points per category.
To track the points properly you have to call 
```
$source = ['QnA', 'CategoryID' => $categoryID];
UserModel::givePoints($userID, $points, $source, $timestamp);
```
where `$categoryID` is the "PointsCategoryID" of the category of the post.

Fortunately there is a convenience method on the CategoryModel that does all the work for you!


DaazKu:
As a side note:

UserModel::givePoints() should never be used in a context where a comment or discussion is involved. CategoryModel::givePoints() should be used instead.

PR #563, opened 2017-11-09
Use a category on the categories page as a link to another category or discussion.

nervoustwit:
Sometimes Admins have want to funnel users who are looking through the categories to another part of the forum or to a specific post. 

This PR creates a feature that will allow admins, when creating or editing a category in the dashboard, to either attach an existing CategoryID or any valid URI that will be used to as the destination whenever a user clicks on or navigates to that category.

### DIFF
--- a/plugins/categoryaslink/class.categoryaslink.plugin.php
+++ b/plugins/categoryaslink/class.categoryaslink.plugin.php
@@ -25,7 +25,7 @@ class CategoryAsLinkPlugin extends Gdn_Plugin {
      */
     public function structure() {
         Gdn::structure()->table('Category')
-            ->column('RedirectURL', 'varchar(255)', true)
+            ->column('RedirectUrl', 'varchar(255)', true)
             ->set();
     }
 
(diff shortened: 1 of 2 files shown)


Charrondev:
I feel like this might be a good idea as a core category type.

plugins/categoryaslink/class.categoryaslink.plugin.php
```diff
@@ -0,0 +1,176 @@
...
+                continue;
+            }
+
+            $linkedCategoryID = val('LinkedCategoryID', $category);
+            if (isset($linkedCategoryID)) {
+                // Get linked category data.
+                $linkedCategory[] = $sender->CategoryModel->categories($linkedCategoryID);
+                $sender->CategoryModel->joinRecent($linkedCategory, $linkedCategoryID);
+                $linkedCategory = $linkedCategory[0];
+                $category['_CssClass'] = 'Aliased AliasedCategory';
```
Charrondev: Couldn't a lot of this be replaced with an assignedment or a clone?

```php
if (is_array($linkedCategory)) {
    $category = $linkedCategory;
} elseif (is_object($linkedCategory) {
    $category = clone $linkedCategory;
}
```

All of the actual values are strings or numbers so they will all get copied by value if you just assign an array, and they can cloned with the clone keyword.

nervoustwit: Thanks, fixed it by just cloning the array. It is never an object. Also, I realized that according to the use case, you create an alias to a category because you want it to have a different name and description.

linc:
Didn't we discuss renaming the field 'LinkedDiscussion'? That doesn't make much sense for a URI that could go anywhere.

nervoustwit:
We can rename the field and tweak the wording but this the forum that needs this done wants to link to discussions in other forums that they own.

linc:
Change the column name to be appropriate.

linc:
We use `Url` not `URL` in database columns.

PR #581, opened 2018-01-11
Fix SQL query pollution in Participated addon

DaazKu:
`DiscussionModel::categoryPermissions()` will query the database if there is no cache.
The plugins was polluting the SQL object before that call.


linc:
What user-facing issue did this cause?

PR #607, opened 2018-04-04
Add logging to JSConnect.

nervoustwit:
Presently JSConnect has no logging. We are basically flying blind troubleshooting on a remote server. This PR will log pertinent information when there is an error during the connection process.


nervoustwit:
This PR got too messy.

PR #608, opened 2018-04-04
Add logging to JSConnect.

nervoustwit:
Presently when you have to troubleshoot JSConnect there is no logging on the remote server. This PR will use our Logger interface to log data whenever there is a problem while connecting.


linc:
I'd be fine with backporting this.

PR #627, opened 2018-04-18
Add the up-to-date discussion's counts to the question record before recalculating it

DaazKu:
By fixing `recalculateDiscussionQnA()` in this [PR](https://github.com/vanilla/addons/pull/620/files#diff-b96a190297fb9e48f8391f75e7da42d3R564) we ended up finding that the discussion record doesn't contains the updated counts on it which makes `recalculateDiscussionQnA()` not assigning the proper `QnA` value in certain cases.


DaazKu:
Properly reference https://github.com/vanilla/addons/pull/620

PR #632, opened 2018-05-29
Bump: Double url() call 

### LINKED ISSUE #631: Bump: Double url() call
## Details

Double URL call causing a problem with Sub-communities when a discussion is bumped.

screen shot 2018-05-29 at 10 26 44 am
screen shot 2018-05-29 at 10 26 50 am

https://github.com/vanilla/addons/blob/4ae429f8dc7b8c0c4097c09f5913f4b090026709/plugins/Bump/class.bump.plugin.php#L16

dannimajdalani:
closes #631 
Double URL call causing a problem with Sub-communities when a discussion is bumped.

### DIFF
--- a/plugins/Bump/class.bump.plugin.php
+++ b/plugins/Bump/class.bump.plugin.php
@@ -13,7 +13,7 @@ class BumpPlugin extends Gdn_Plugin {
         $discussion = $args['Discussion'];
         if (checkPermission('Garden.Moderation.Manage')) {
             $label = t('Bump');
-            $url = "discussion/bump?discussionid={$discussion->DiscussionID}";
+            $url = url("discussion/bump?discussionid={$discussion->DiscussionID}", true);
             // Deal with inconsistencies in how options are passed
             if (isset($sender->Options)) {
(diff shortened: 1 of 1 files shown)


plugins/Bump/class.bump.plugin.php
```diff
@@ -13,7 +13,7 @@ public function base_discussionOptions_handler($sender, $args) {
         $discussion = $args['Discussion'];
         if (checkPermission('Garden.Moderation.Manage')) {
             $label = t('Bump');
-            $url = url("discussion/bump?discussionid={$discussion->DiscussionID}");
+            $url = "discussion/bump?discussionid={$discussion->DiscussionID}";
```
charrondev: By not wrapping this at all you may be causing a problem downstream, when a totally unqualified url gets set in the Args

https://github.com/vanilla/addons/blob/69468c31e252978ca16e58d8d32d463fcf1f3a7f/plugins/Bump/class.bump.plugin.php#L23

The better solution here is likely to still use the `url()` function, but to pass true as the second parameter so it becomes an absolute url. This will prevent subsequent calls from breaking the url.

PR #639, opened 2018-06-27
Fix the Category as Link Addon so it works with Subcommunities.

nervoustwit:
This PR fixes the fact that the Category as Link plugin doesn't work when subcommunities are turned on by doing two things:

 1. Making the Category as Link fire first (changing the priority)
 2. Using the Subcommunities way of constructing the URL if Subcommunities are turned on.

### DIFF
--- a/plugins/categoryaslink/class.categoryaslink.plugin.php
+++ b/plugins/categoryaslink/class.categoryaslink.plugin.php
@@ -122,9 +122,8 @@ if (!function_exists("categoryUrl")) {
             return safeURL(val('RedirectUrl', $category));
         }
 
-        // Build URL.
+        // SEOLinks version.
         if (class_exists('SEOLinksPlugin')) {
-            // SEOLinks version
             if (!isset($px)) {
                 $px = SEOLinksPlugin::Prefix();
             }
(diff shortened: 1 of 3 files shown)


linc:
It's a rule: You can't change the priority level of an addon without adding docs about why.

linc:
This is fine after the docs are added.

plugins/categoryaslink/class.categoryaslink.plugin.php
```diff
@@ -122,22 +122,28 @@ function categoryUrl($category, $page = '', $withDomain = true) {
...
-        } else {
-            // Normal version
-            $result = '/categories/'.rawurlencode($category['UrlCode']);
-            if ($page && $page > 1) {
-                $result .= '/p'.$page;
-            }
+            return $result;
+        }
+
+        // Subcommunities version.
```
initvector: Should this come before SEO Links?

plugins/categoryaslink/class.categoryaslink.plugin.php
```diff
@@ -122,22 +122,28 @@ function categoryUrl($category, $page = '', $withDomain = true) {
             return safeURL(val('RedirectUrl', $category));
         }
 
-        // Build URL.
```
initvector: It looks like we went from having one if/else statement and one return statement to two separate if statements and three returns, which is maybe needlessly increasing the complexity here. For the sake of readability, I think you might be able to simplify "building the URL" into a single if/else-if/else statement and one return.

PR #642, opened 2018-07-11
Backport: Fix the Category as Link Addon so it works with Subcommunities.

nervoustwit:
Backport PR #639 

This PR fixes the fact that the Category as Link plugin doesn't work when subcommunities are turned on by doing two things:

Making the Category as Link fire first (changing the priority)
Using the Subcommunities way of constructing the URL if Subcommunities are turned on.


initvector:
After more than a week, I'm gonna assume it's a "yes" and add 'er in.

Issue #643, opened 2018-07-16
Error in QnA question/discussion state update logic


topcheese13:
The problem
---
https://github.com/vanilla/addons/blob/master/plugins/QnA/class.qna.plugin.php#L557

In the QnA plugin we have a function named `recalculateDiscussionQnA` that gets called in multiple places in the plugin. For our case it gets call by this hook `commentModel_beforeUpdateCommentCount_handler` after somebody comments.

The logic of this function is faulty. 

```
$row = Gdn::sql()->getWhere('Comment',
            ['DiscussionID' => val('DiscussionID', $discussion), 'QnA is not null' => ''], 'QnA, DateAccepted', 'asc', 1)->firstRow(DATASET_TYPE_ARRAY);
```

The first step of the function is the above, we do a query to look for the first comment that matches our discussionID, that as a current QnA state of not null (the only possible states are NULL, Accepted and Rejected).

Let's say for the sake of our example that one answer has been posted and it has been marked as rejected, than a new answer is posted. By our query parameters, the rejected comment will be returned as `$row`.

From there we start the parsing of our returned row.

```
if (!$row) {
            if (val('CountComments', $discussion, 0) > 0) {
                $set['QnA'] = 'Answered';
            } else {
                $set['QnA'] = 'Unanswered';
            }

            $set['DateAccepted'] = null;
            $set['DateOfAnswer'] = null;
        } elseif ($row['QnA'] == 'Accepted') {
            $set['QnA'] = 'Accepted';
            $set['DateAccepted'] = $row['DateAccepted'];
            $set['DateOfAnswer'] = $row['DateInserted'];
        } elseif ($row['QnA'] == 'Rejected') {
            $set['QnA'] = 'Rejected';
            $set['DateAccepted'] = null;
            $set['DateOfAnswer'] = null;
        }
```

From this we are setting (or leaving as if you prefer) the QnA status of the question to Rejected , when it should be updated to Answered.

Info
---

Questions have 4 possible states `'Unanswered','Answered','Accepted','Rejected'`
Answers have 3 possible states `'Accepted','Rejected'` or NULL. Null means the answer hasn't been treated yet.

The fix
---
We have to review our logic. I would "recalculateDiscussionQnA" this way.

1. If we have any comment marked as Accepted. The question should be marked as Accepted. If we don't go to 2.

2. If we have any comment marked as NULL. The question should be marked as Answered. If we don't go to 3.

[...]

PR #644, opened 2018-07-26
Fix recalculation of discussions QnA state

topcheese13:
Closes https://github.com/vanilla/addons/issues/643.

Refactored the `recalculateDiscussionQnA()` function based on the logic established in https://github.com/vanilla/addons/issues/643.

Also, removed the faulty separate logic to set the discussion's QnA state when an answer is treated as helpful or not, now using the refactored `recalculateDiscussionQnA()`. The discussion state is now being set to the correct state regardless of how many answers remain to be treated.

### DIFF
--- a/plugins/QnA/class.qna.plugin.php
+++ b/plugins/QnA/class.qna.plugin.php
@@ -554,8 +554,9 @@ class QnAPlugin extends Gdn_Plugin {
         $set = [];
 
         // Look for at least one accepted answer/comment.
-        $acceptedComment = Gdn::sql()->getWhere('Comment',
-            ['DiscussionID' => val('DiscussionID', $discussion), 'QnA' => 'Accepted'], '', 'asc', 1)->firstRow(DATASET_TYPE_ARRAY);
+        $acceptedComment = Gdn::sql()->getWhere(
+            'Comment', ['DiscussionID' => val('DiscussionID', $discussion), 'QnA' => 'Accepted'], '', 'asc', 1
+        )->firstRow(DATASET_TYPE_ARRAY);
 
         if ($acceptedComment) {
             $set['QnA'] = 'Accepted';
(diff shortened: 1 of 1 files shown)


DaazKu:
I highly suggest that we add an integration test for this one since that recalculate function has been refactored multiple times already.

charrondev:
Just to add a bit of context here https://github.com/vanilla/addons/pull/443

plugins/QnA/class.qna.plugin.php
```diff
@@ -553,31 +548,36 @@ public function discussionController_qnAOptions_create($sender, $discussionID =
...
+            $set['DateAccepted'] = $acceptedComment['DateAccepted'];
+            $set['DateOfAnswer'] = $acceptedComment['DateInserted'];
+        } else {
+            // Look for at least one untreated answer/comment.
+            $answeredCommment = Gdn::sql()->getWhere('Comment',
+                ['DiscussionID' => val('DiscussionID', $discussion), 'QnA is null' => ''], '', 'asc', 1)->firstRow(DATASET_TYPE_ARRAY);
+            if ($answeredCommment) {
                 $set['QnA'] = 'Answered';
+                $set['DateAccepted'] = null;
+                $set['DateOfAnswer'] = null;
```
charrondev: If the comment was answered should we be setting the DateOfAnswer here?

plugins/QnA/class.qna.plugin.php
```diff
@@ -553,31 +548,36 @@ public function discussionController_qnAOptions_create($sender, $discussionID =
...
+            } else if (val('CountComments', $discussion, 0) > 0) {
```
charrondev: Could we pull this condition out into it's own variable, maybe `$hasComments`?

plugins/QnA/class.qna.plugin.php
topcheese13: No we shouldn't. I removed it in the lastest commit.

plugins/QnA/class.qna.plugin.php
topcheese13: Added this to the latest commit as well.

PR #649, opened 2018-08-10
Fix broken SSO in Steam Connect

### LINKED ISSUE #647: SteamConnect "UniqueID is required"
image

Connecting steam is just pasting your API key into the settings, so theres not much a user could do to debug. I replicated this on my demo: 
image

The steam app gives no feedback on what might be wrong, and we have no instructions on what needs to go on into the "domain" settings on the stream side. I revoked my API key and got the same error. I put in a new one and got the same error.

dannimajdalani:
closes #647 

The error "uniqueID is required" was because we were trying to match `http://steamcommunity.com/.. `with the received` https://steamcommunity.com/.. `

[class.steamconnect.plugin.php#L125](https://github.com/vanilla/addons/blob/master/plugins/SteamConnect/class.steamconnect.plugin.php#L125 )

### DIFF
--- a/plugins/SteamConnect/class.steamconnect.plugin.php
+++ b/plugins/SteamConnect/class.steamconnect.plugin.php
@@ -12,37 +12,49 @@ class SteamConnectPlugin extends Gdn_Plugin {
     /**
      * This will run when you "Enable" the plugin
      *
-     * @since  1.0.0
-     * @access public
      * @return bool
      */
     public function setup() {
 
     }
 
+    /**
+     * Check whether this addon has enough configuration to work.
+     *
+     * @return mixed
+     */
     public function isConfig() {
         return c('Plugins.SteamConnect.APIKey', FALSE);
     }
 
+    /**
+     * Retrieve the URL to start an auth request.
+     *
+     * @param bool $popup
+     * @return string
+     */
     protected function _AuthorizeHref($popup = FALSE) {
         $url = url('/entry/openid', TRUE);
         $urlParts = explode('?', $url);
         parse_str(getValue(1, $urlParts, ''), $query);
-        $query['url'] = 'http://steamcommunity.com/openid';
+        $query['url'] = 'https://steamcommunity.com/openid';
         $path = '/'.Gdn::request()->path();
         $query['Target'] = getValue('Target', $_GET, $path ? $path : '/');
-        if ($popup)
+        if ($popup) {
             $query['display'] = 'popup';
+        }
 
         $result = $urlParts[0].'?'.http_build_query($query);
         return $result;
     }
 
-
-    /// Plugin Event Handlers ///
-
+    /**
+     * Add Steam as an option to the normal signin page.
+     *
+     * @param $sender
+     * @param $args
+     */
     public function entryController_signIn_handler($sender, $args) {
-
         if (isset($sender->Data['Methods']) && $this->isConfig()) {
             $url = $this->_AuthorizeHref();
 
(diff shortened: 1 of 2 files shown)


cchabilall83:
Most of the DocBlock are missing some information.  I didn't comment on all of them, but you can see the pattern from the few I did.

initvector:
> Most of the DocBlock are missing some information. I didn't comment on all of them, but you can see the pattern from the few I did.

I don't see any comments. ☹️

plugins/SteamConnect/class.steamconnect.plugin.php
```diff
@@ -12,37 +12,49 @@ class SteamConnectPlugin extends Gdn_Plugin {
...
-
-    /// Plugin Event Handlers ///
-
+    /**
+     * Add Steam as an option to the normal signin page.
+     *
+     * @param $sender
+     * @param $args
+     */
     public function entryController_signIn_handler($sender, $args) {
```
initvector: Could use some types on these parameter tags.

plugins/SteamConnect/class.steamconnect.plugin.php
```diff
@@ -56,37 +68,69 @@ public function entryController_signIn_handler($sender, $args) {
...
         if ($this->isConfig()) {
             echo "\n".$this->_GetButton();
         }
     }
 
+    /**
+     * Inject a sign-in icon into the ME menu.
+     *
+     * @param Gdn_Controller $sender.
+     * @param Gdn_Controller $args.
```
initvector: Is `$args` supposed to be an instance of `Gdn_Controller`?

plugins/SteamConnect/class.steamconnect.plugin.php
```diff
@@ -56,37 +68,69 @@ public function entryController_signIn_handler($sender, $args) {
...
+     * @param AssetModel $sender
+     */
     public function assetModel_styleCss_handler($sender) {
         $sender->addCssFile('steam-connect.css', 'plugins/SteamConnect');
     }
 
+    /**
+     * Build-A-Button.
+     *
+     * @return string
```
initvector: It looks like this could potentially return `null` (nothing).

plugins/SteamConnect/class.steamconnect.plugin.php
```diff
@@ -56,37 +68,69 @@ public function entryController_signIn_handler($sender, $args) {
...
         if ($this->isConfig()) {
             $url = $this->_AuthorizeHref();
             return socialSigninButton('Steam', $url, 'icon', ['class' => 'js-extern']);
         }
     }
 
+    /**
+     * Add Steam-Connect button to default mobile theme.
+     *
+     * @param $sender
```
initvector: What type of value would `$sender` be?

plugins/SteamConnect/class.steamconnect.plugin.php
```diff
@@ -56,37 +68,69 @@ public function entryController_signIn_handler($sender, $args) {
...
         if (!Gdn::session()->isValid() && $this->isConfig()) {
             echo "\n".wrap($this->_GetButton(), 'li', ['class' => 'Connect SteamConnect']);
         }
     }
 
+    /**
+     * Capture the user's UniqueID being sent back from provider and saving it to the session.
+     *
+     * @param $sender
+     * @param $args
```
initvector: I think we can add types for these parameters.

plugins/SteamConnect/class.steamconnect.plugin.php
```diff
@@ -121,18 +165,29 @@ public function openIDPlugin_afterConnectData_handler($sender, $args) {
         }
     }
 
+    /**
+     * Get steam id
+     *
+     * @param $openID
```
initvector: What type of value would `$openID` be?

plugins/SteamConnect/class.steamconnect.plugin.php
```diff
@@ -121,18 +165,29 @@ public function openIDPlugin_afterConnectData_handler($sender, $args) {
...
+     * @return mixed
```
initvector: "mixed" would typically imply you can't be sure what's returned, but it looks like there are only a couple of possibilities here.

PR #650, opened 2018-08-20
Add test to cover questions after an answer has been rejected

### LINKED ISSUE #646: Expand CommentsAnswerTest to cover the case from #643
We should expand the `CommentsAnswerTest` to cover the issue outlined in https://github.com/vanilla/addons/issues/643.

## Success Criteria
- The tests should be updated such that reverting https://github.com/vanilla/addons/issues/643 causes them to fail.
- The tests should then pass with https://github.com/vanilla/addons/issues/643 applied.
- Tests should cover the now 4 different states `Accepted`, `Answered`, `Rejected`, `Unanswered`.
- ~~Tests should also cover `DateAccepted` and `DateOfAnswer` if those or lowercased equivalents get passed through the API.~~

cchabilall83:
This test that after a question has an rejected answer then another answer is posted that the question's status will be **answered**.  An additional test to verify that the status of a posted question 
is **unanswered**.  Tests to validate the date accepted and date answered fields were also added

Closed #646

### DIFF
--- a/plugins/QnA/tests/APIv2/CommentsAnswerTest.php
+++ b/plugins/QnA/tests/APIv2/CommentsAnswerTest.php
@@ -211,4 +211,43 @@ class CommentsAnswerTest extends AbstractAPIv2Test {
         $updatedQuestion = $this->getQuestion($question['discussionID']);
         $this->assertIsQuestion($updatedQuestion, ['status' => 'rejected']);
     }
+
+    /**
+     * Test rejecting an answer and then resubmitting an answer.
+     *
+     * @depends testPostAnswer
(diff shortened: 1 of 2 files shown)


initvector:
I always forget to comment in the body of the review, so now I'll just have to amend it here!

> I didn't add a test to cover the 4th point to test datas, it should be filtered by our schema, so I though it would be redundant. Correct me if i'm wrong.

It looks like those dates (`dateAccepted` and `dateOfAnswer`) are part of the API output. If we don't have tests making sure they're properly set at relevant times, we definitely should.

PR #655, opened 2018-09-12
Add cache options to SQL driver to cache getProviders query response 

alex-mtl:
Add cache options to SQL driver to cache getProviders query response for 15 min.
That allows app to avoid unnecessary DB connection to be initialized in some cases.

I've discussed this patch with @DaazKu. 
His opinion - we should provide wider solution with some global scope but not only apply to jsconnect plugin.

Fixes: https://github.com/vanilla/vanilla-patches/issues/387

### DIFF
--- a/plugins/jsconnect/class.jsconnect.plugin.php
+++ b/plugins/jsconnect/class.jsconnect.plugin.php
@@ -207,13 +207,19 @@ class JsConnectPlugin extends Gdn_Plugin {
      * @return array|mixed
      */
     public static function getProvider($client_id = null) {
+        $sqlCacheKey = 'getProvider:';
         if ($client_id !== null) {
+            $sqlCacheKey .= 'AuthenticationKey:'.$client_id;
             $where = ['AuthenticationKey' => $client_id];
         } else {
+            $sqlCacheKey .= 'AuthenticationSchemeAlias:jsconnect';
             $where = ['AuthenticationSchemeAlias' => 'jsconnect'];
         }
 
-        $result = Gdn::sql()->getWhere('UserAuthenticationProvider', $where)->resultArray();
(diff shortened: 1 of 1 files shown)


DaazKu:
So I want to provide some context here:

We were in an impasse on Ops side so I tried to get this patched ASAP but at the "last minute" (1h ago) CloudFlare unlocked a new BETA feature on our account that allowed us to mitigate the issue.
Since it is not something that needs to be fixed ASAP anymore the product fix should be global and well planned. (Read sprinted)

DaazKu:
I updated the issue to make things clear but the whole Gdn_AuthenticationProvider model needs to be cached. It's a widespread problem that does not only hit jsConnect.

plugins/jsconnect/class.jsconnect.plugin.php
```diff
@@ -676,6 +689,9 @@ protected function settings_addEdit($sender, $args) {
         $sender->addJsFile('jsconnect-settings.js', 'plugins/jsconnect');
 
         $client_id = $sender->Request->get('client_id');
+
+        Gdn::cache()->remove(self::getProviderSqlCacheKey($client_id));
```
initvector: This is maybe a little premature. I think you're going to be purging the cache whenever the settings page is opened, instead of when the provider is actually updated. What do you think about moving this down a few lines, after a successful save?

```php
if ($form->save(['ID' => $client_id])) {
    Gdn::cache()->remove(self::getProviderSqlCacheKey($client_id));
    $sender->setRedirectTo('/settings/jsconnect');
}
```

alex-mtl: Ok. Done.

Issue #74, opened 2014-04-07
QnA required version is incorrect


DataTables:
The master QnA plug-in file currently says that v2.0.18.0 or newer is required, however, it doesn't work in 2.1b3, I believe due to this commit: https://github.com/vanillaforums/Addons/commit/fb6d54211af54

Specifically `Base_DiscussionTypes_Handler` isn't run in 2.0 or 2.1, so I presume it is in 2.2? The required version should be bumped to reflect this.

PR #660, opened 2018-10-05
User option to disable/enable QnA notifications

dannimajdalani:
Closes [#3](https://github.com/vanilla/support/issues/3)

**What Changed**

I added notification preferences for QnA when:

- A question is answered 
- An answer is accepted

### DIFF
--- a/plugins/QnA/class.qna.plugin.php
+++ b/plugins/QnA/class.qna.plugin.php
@@ -1385,9 +1385,9 @@ class QnAPlugin extends Gdn_Plugin {
      * @param ProfileController $sender
      */
     public function profileController_afterPreferencesDefined_handler($sender) {
-        $sender->Preferences['Notifications']['Email.AnswerAccepted'] = t('Notify me when people accept my answer');
-        $sender->Preferences['Notifications']['Popup.AnswerAccepted'] = t('Notify me when people accept my answer');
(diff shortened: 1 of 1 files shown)


cchabilall83:
I tested every scenario, it seemed to work.  I was getting some weird results, but I think I wasn't saving my preferences properly. @topcheese13 if you receive any inconsistent results let me know.

cchabilall83:
also do the strings need to be translated?

dannimajdalani:
> also do the strings need to be translated?

Yes. They are using the translation function, but I will create a separate PR for the keys.

dannimajdalani:
Translation PR [#74](https://github.com/vanilla/locales/pull/74)

plugins/QnA/class.qna.plugin.php
```diff
@@ -1378,4 +1378,16 @@ public function searchResultSchema_init(Schema $schema) {
         $types[] = 'question';
         $schema->setField('properties.type.enum', $types);
     }
+
+    /**
+     * Adds status notification options to profiles.
+     *
```
cchabilall83: this doesn't look right

dannimajdalani: This is a description of what follows.

PR #661, opened 2018-10-10
Add question states recalculation to dba/counts

topcheese13:
The goal here was to add an option to recalculate the current QnA state of all questions through the `dba/counts` endpoint.

Relevant information on the PR
---
* Recalculation happens in batches of 1000 questions at a time to prevent locking the table for too long.
* There is no "capping" on how many questions you can update. (Should we add one?)
* I tried optimizing queries as much as possible, as an example 100,000 questions with 200,000 answers take about 2 minutes to fully update on my local.
* On the UI you can see a "progress" in percentage to preventing feeling like the job is stuck.
* You can refer to the `recalculateDiscussionsQnA()` method to see the logic I used to update the records.

### DIFF
--- a/plugins/QnA/models/class.qnamodel.php
+++ b/plugins/QnA/models/class.qnamodel.php
@@ -12,7 +12,7 @@ class QnaModel extends Gdn_Model {
     public function counts($column) {
         $result = ['Complete' => true];
 
-        switch($column) {
+        switch ($column) {
             // Discussion table, QnA column will be updated.
             case 'QnA':
                 $request = Gdn::request()->get();
@@ -105,13 +105,13 @@ class QnaModel extends Gdn_Model {
 
         $currentBatch = array_column($currentBatch, 'DiscussionID', 'DiscussionID');
 
-        $latestID = key(array_slice($currentBatch,-1,1,TRUE));
+        $latestID = key(array_slice($currentBatch, -1, 1, true));
 
         $this->recalculateDiscussionsQnA($currentBatch);
 
         $numberOfBatchesDone++;
 
-        if ($totalBatches == $numberOfBatchesDone) {
+        if ($totalBatches === $numberOfBatchesDone) {
             return ['Complete' => true];
         }
 
--- a/plugins/QnA/models/class.qnamodel.php
+++ b/plugins/QnA/models/class.qnamodel.php
@@ -1,10 +1,5 @@
 <?php
 /**
- * @copyright 2009-2018 Vanilla Forums Inc.
- * @license Proprietary
- */
-
-/**
  * Class QnaModel
  */
 class QnaModel extends Gdn_Model {
--- a/plugins/QnA/models/class.qnamodel.php
+++ b/plugins/QnA/models/class.qnamodel.php
@@ -111,7 +111,7 @@ class QnaModel extends Gdn_Model {
 
         $numberOfBatchesDone++;
 
-        if ($totalBatches === $numberOfBatchesDone) {
+        if ($totalBatches == $numberOfBatchesDone) {
             return ['Complete' => true];
         }
 
--- a/plugins/QnA/models/class.qnamodel.php
+++ b/plugins/QnA/models/class.qnamodel.php
@@ -1,0 +1,131 @@
+<?php
+/**
+ * @copyright 2009-2018 Vanilla Forums Inc.
+ * @license Proprietary
+ */
+
+/**
+ * Class QnaModel
+ */
+class QnaModel extends Gdn_Model {
+    /**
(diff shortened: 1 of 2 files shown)


plugins/QnA/models/class.qnamodel.php
```diff
@@ -0,0 +1,126 @@
...
+    /**
+     * Preparing batches of discussions (questions) and triggering recalculation of their QnA state.
+     * Updating records in batches of 1000 to make sure we don't lock the table for too long.
+     *
+     * @param integer $numberOfBatchesDone number of batches already processed
+     * @param integer $latestID latest discussionID we treated
+     * @return array current state of QnA recalculation. Formatted to match what "dba.js" expects
+     */
+    private function recalculateDiscussionQnABatches($numberOfBatchesDone, $latestID) {
+        $perBatch = 1000;
```
cchabilall83: I think this could be a constant.

plugins/QnA/models/class.qnamodel.php
```diff
@@ -0,0 +1,126 @@
...
+            ])
+            ->orderBy('DiscussionID')
+            ->limit($perBatch)
+            ->get()
+            ->resultArray();
+
+        $currentBatch = array_column($currentBatch, 'DiscussionID', 'DiscussionID');
+
+        $latestID = key(array_slice($currentBatch,-1,1,TRUE));
+
```
cchabilall83: Spacing between the commas, and TRUE should be lowercase

plugins/QnA/models/class.qnamodel.php
```diff
@@ -0,0 +1,126 @@
...
+     * Update Q&A counts through the dba/counts endpoint.
+     *
+     * @param string $column
+     * @return array $results Formatted to match what "dba.js" expects
+     */
+    public function counts($column) {
+        $result = ['Complete' => true];
+
+        switch($column) {
+            // Discussion table, QnA column will be updated.
```
cchabilall83: space after the switch

plugins/QnA/models/class.qnamodel.php
```diff
@@ -0,0 +1,126 @@
...
+            case 'QnA':
+                $request = Gdn::request()->get();
+                $result = $this->recalculateDiscussionQnABatches($request['NumberOfBatchesDone'] ?? 0, $request['LatestID'] ?? 0);
+                break;
+        }
```
cchabilall83: not sure if the break is lined up properly.  In my IDE it's line up with the case, it should be lined up with $result if i'm not mistaken.

cchabilall83: No default needed?  Throw an Exception?

plugins/QnA/models/class.qnamodel.php
```diff
@@ -0,0 +1,126 @@
...
+        $currentBatch = array_column($currentBatch, 'DiscussionID', 'DiscussionID');
+
+        $latestID = key(array_slice($currentBatch,-1,1,TRUE));
+
+        $this->recalculateDiscussionsQnA($currentBatch);
+
+        $numberOfBatchesDone++;
+
+        if ($totalBatches == $numberOfBatchesDone) {
+            return ['Complete' => true];
```
cchabilall83: shouldn't it be === just incase?

plugins/QnA/models/class.qnamodel.php
topcheese13: I thought about this, is there a gain to make it a constant if it's not used anywhere else ?

plugins/QnA/models/class.qnamodel.php
topcheese13: fixed

plugins/QnA/models/class.qnamodel.php
topcheese13: fixed

plugins/QnA/models/class.qnamodel.php
topcheese13: It's lined up with $result

topcheese13: I added the switch because of how other counts work around Vanilla. It's also easier to add other counts to QnA if we need to in the future.

plugins/QnA/models/class.qnamodel.php
topcheese13: I didn't want to make it to restrictive, especially since one is received from the ajax request and the other one is the result of an ajax query but it could make more sense to triple equal this. Done.

cchabilall83: really?  what data types are you comparing?

topcheese13: http://php.net/ceil

float vs integer

PR #663, opened 2018-10-22
Create a plugin that will set the default controller canonical tag.

nervoustwit:
This PR will add an addon that will make the Default Controller (or "Home Page") only be accessible via the Document Root ("/") of the forum for Search Engine Optimization purposes, and make the canonical tag reference the domain name only (no path).

### DIFF
--- a/plugins/homepagecanonical/class.homepagecanonical.plugin.php
+++ b/plugins/homepagecanonical/class.homepagecanonical.plugin.php
@@ -1,0 +1,32 @@
+<?php
+/**
+ * Homepage Canonical Plugin.
+ *
+ * @author Patrick Kelly <patrick.k@vanillaforums.com>
+ * @copyright 2009-2018 Vanilla Forums Inc.
+ * @license GNU GPLv2 http://www.opensource.org/licenses/gpl-2.0.php
+ */
+
+class HomepageCanonicalPlugin extends Gdn_Plugin {
+
+    /**
+     * If the Default Controller page is requested (Discussions or Categories), 301 redirect to "/" and insert the Canonical tag.
+     *
+     * @param VanillaController $sender
+     * @param array $args
+     */
+    public function base_render_before($sender, $args) {
+        // Use $SERVER['REQUEST_URI'], and not request()->path() because we alter path;
+        $requestURI = Gdn::request()->getValueFrom('server', 'REQUEST_URI');
+        $defaultController = Gdn::router()->getDestination('DefaultController');
+        if ((strpos($requestURI,'?') === false) && preg_match('/\/'.$defaultController.'\/?$/i',$requestURI)) {
+            safeHeader('Location: '.url('/', true), true, '301');
+        }
+
+        if ($requestURI === '/') {
+            saveToConfig('Garden.Modules.NoCanonicalUrl', true, false);
+            $sender->Head->addTag('link', ['rel' => 'canonical', 'href' => url('/', true)], null, false);
+        }
+    }
+
+}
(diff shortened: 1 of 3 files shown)


plugins/homepagecanonical/class.homepagecanonical.plugin.php
```diff
@@ -0,0 +1,32 @@
...
+
+    /**
+     * If the Default Controller page is requested (Discussions or Categories), 301 redirect to "/" and insert the Canonical tag.
+     *
+     * @param VanillaController $sender
+     * @param array $args
+     */
+    public function base_render_before($sender, $args) {
+        // Use $SERVER['REQUEST_URI'], and not request()->path() because we alter path;
+        $requestURI = Gdn::request()->getValueFrom('server', 'REQUEST_URI');
```
linc: `getValueFrom()` is essentially `val()` for the request object. Elsewhere I see us using `Gdn::request()->getRequestArguments('server')['REQUEST_URI']` so I'd roll with that.

plugins/homepagecanonical/class.homepagecanonical.plugin.php
```diff
@@ -0,0 +1,32 @@
...
+
+class HomepageCanonicalPlugin extends Gdn_Plugin {
+
+    /**
+     * If the Default Controller page is requested (Discussions or Categories), 301 redirect to "/" and insert the Canonical tag.
+     *
+     * @param VanillaController $sender
+     * @param array $args
+     */
+    public function base_render_before($sender, $args) {
```
linc: I believe this hook is too late for wide adoption. Is there an earlier hook we can safely evaluate the URL to avoid needless database querying & page building before doing a redirect?

plugins/homepagecanonical/class.homepagecanonical.plugin.php
```diff
@@ -0,0 +1,32 @@
...
+        // Use $SERVER['REQUEST_URI'], and not request()->path() because we alter path;
+        $requestURI = Gdn::request()->getValueFrom('server', 'REQUEST_URI');
+        $defaultController = Gdn::router()->getDestination('DefaultController');
+        if ((strpos($requestURI,'?') === false) && preg_match('/\/'.$defaultController.'\/?$/i',$requestURI)) {
+            safeHeader('Location: '.url('/', true), true, '301');
+        }
+
+        if ($requestURI === '/') {
+            saveToConfig('Garden.Modules.NoCanonicalUrl', true, false);
```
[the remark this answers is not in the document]
linc: I was going to say it at least needs a line of docs to explain itself.

linc:
> Is there a particular reason that this is a being implemented as a plugin and not directly in core? This URL structure doesn't doesn't seem that out there to me. I'd rather see a config option somewhere in the dashboard enabling this.

It's SEO voodoo, forever-301s lead to problems, and we want a [real homepage](https://github.com/vanilla/vanilla/issues/4908) that this could conflict with.

This addon exists because humans can't be reasoned with, not because it's a good feature.

plugins/homepagecanonical/class.homepagecanonical.plugin.php
nervoustwit: I split it into to methods so that the redirecting happens in the dispatcher function.

nervoustwit:
Ten comments on 50 lines of code. But on the plus side, I didn't label it as a Gimme.

plugins/homepagecanonical/class.homepagecanonical.plugin.php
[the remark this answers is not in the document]
nervoustwit: This line has been moved and documented.

PR #687, opened 2019-02-11
Update licenses and versions for open source release

linc:
These changes reflect what I released on the open source directory for these addons on Sunday, Feb 10. It consists of version bumps, license additions, and min requirement increases (usually to 2.5 to be safe).


linc:
This should also go to `release/2.8` since it reflects the releases we did concurrent with that version.

PR #695, opened 2019-03-07
Allow add-ons to customize the robots.txt

tburry:
This change makes a `Robots` data model that can be modified by add-ons to add sitemaps and add rules to the robots.txt file. As we add more applications we need a way for them to add their sitemaps to the application and this is meant to accomplish that.


tburry:
As an FYI, this PR is a result of me fleshing out the requirements for KB sitemaps. Probably doesn't need to be approved this sprint if it's too much.

plugins/Sitemaps/Robots.php
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * Get the list of robots rules.
+     *
+     * @return array Returns an array of rules.
+     */
+    public function getRules(): array {
+        return $this->rules;
+    }
+}
```
charrondev: Missing a line ending here.

plugins/Sitemaps/class.sitemaps.plugin.php
```diff
@@ -110,7 +110,7 @@ public function utilityController_robots_create($sender) {
         $sender->deliveryType(DELIVERY_TYPE_VIEW);
         $sender->setHeader('Content-Type', 'text/plain');
 
-        $robots = new \Vanilla\Sitemaps\Robots();
+        $robots = new \Vanilla\Sitemap\Robots();
```
tburry: The namespace really should match the name of the addon which is plural.

tburry:
WRT moving the functionality into core I agree with the sentiment, but I think that's too much yak shaving for this sentiment. Keep in mind that kb doesn't really depend on sitemaps to work. This would be more like a "suggested addon".

plugins/Sitemaps/class.sitemaps.plugin.php
charrondev: In any event, we've moved this class up into core now. I think the class is now `Vanilla\Web\Robots`.

PR #696, opened 2019-03-21
Add expanding QnA answers in discussions API resource

### LINKED ISSUE #676: Add answer ID to discussion API response for questions
The QnA add-on adds attributes to discussions with a "question" type in responses from the discussions API. It currently includes `status`, `dateAccepted` and `dateAnswered`. Details about the answer should be included (eg. comment ID, maybe expanding the content of the post), if available. Multiple accepted answers are a possibility that needs to be considered.

initvector:
This update adds including accepted answers as part of the "question" attributes on a discussion. By default, only the comment IDs of accepted answers come down. However, users may use "acceptedAnswers" in the "expand" parameter of a call to /api/v2/discussions/{id}, which will return the body of those comments, as well.

Closes #676

### DIFF
--- a/plugins/QnA/class.qna.plugin.php
+++ b/plugins/QnA/class.qna.plugin.php
@@ -1189,6 +1189,20 @@ class QnAPlugin extends Gdn_Plugin {
             ],
             'dateAccepted:dt|n' => 'When an answer was accepted.',
             'dateAnswered:dt|n' => 'When the last answer was inserted.',
+            "acceptedAnswers" => [
+                "object" => "Accepted answers.",
+                "type" => "array",
+                "items" => Schema::parse([
+                    "commentID" => [
+                        "type" => "integer",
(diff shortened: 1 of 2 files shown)


plugins/QnA/class.qna.plugin.php
```diff
@@ -1189,6 +1189,20 @@ public function fullQuestionMetaDataSchema() {
...
             'dateAnswered:dt|n' => 'When the last answer was inserted.',
+            "acceptedAnswers" => [
+                "object" => "Accepted answers.",
+                "type" => "array",
+                "items" => Schema::parse([
+                    "commentID" => [
+                        "type" => "integer",
+                        "description" => "Unique ID of the accepted answer's comment row."
+                    ],
+                    "body?" => [
```
alex-mtl: Are you sure we can make body optional?

PR #711, opened 2019-05-25
Fix conflict with Bump plugin

pioc92:
If you are using NoIndex addon and Bump addon, you can't use bump if you enable it after NoIndex because NoIndex plugin has the same attribute


charrondev:
I'm going to try to get this reviewed early next week.

PR #712, opened 2019-05-25
Fix key problem

pioc92:
The folder of the addon is Bump and the key is bump, so there is an error

### DIFF
--- a/plugins/Bump/addon.json
+++ b/plugins/Bump/addon.json
@@ -3,7 +3,7 @@
     "description": "Allows moderators to bump a discussion without commenting.",
     "version": "1.0.1",
     "mobileFriendly": true,
-    "key": "bump",
+    "key": "Bump",
     "type": "addon",
     "authors": [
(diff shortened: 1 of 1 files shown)


charrondev:
Sorry, but this is a bit of a no-go. It's a actually a huge job to migrate all of the configurations on our infrastructure to use the new configuration and time it properly with the release (and also for people in the open source community).

I'm not saying it can't be done, but it would be more consistent and also a lot easier to the other way.

- Our naming convention for addon keys/directories is [lower-dashed-case going forward](https://docs.vanillaforums.com/developer/addons/addon-info/#key).
- Making the directory name match the addon key means no configuration migration will be necessary.

When I started at vanilla a couple years ago I wanted to rename a lot of things to make them more consistent, but I will say it's a ton of work to successfully rename something everywhere (across our infrastructure and open source). Case-sensitive vs. Case-insensitive systems can play a role as well.

PR #714, opened 2019-05-30
Add warning to sitemaps plugin when private communities is enabled.

charrondev:
I had this sitting around on my localhost. I had forgotten I had done, but now is as good a time as any.

- Adds a warning to the sitemaps page if private communities is enabled.
- Converts the view to a twig view.

_Note: These strings were not translated, and I'm not considering it as part of the scope of this to translate them._


plugins/Sitemaps/views/settings.twig
```diff
@@ -0,0 +1,33 @@
+<h1>{{Title}}</h1>
+<div class="padded">
+    {% if %}
```
initvector: 🤔

plugins/Sitemaps/views/settings.twig
```diff
@@ -0,0 +1,33 @@
...
+        </li>
+    </ul>
+    </p>
+    <p>{{t('Learn more about sitemaps from the following sites:')}}</p>
+    <ul>
+        <li>
+            <a href="https://www.sitemaps.org/" target="_blank" rel="noopener">sitemaps.org</a>
+        </li>
+        <li>
+            <a href="https://www.google.com/webmasters/" target="_blank" rel="nooopener">Google webmaster center</a>
```
initvector:
```suggestion
            <a href="https://www.google.com/webmasters/" target="_blank" rel="noopener">Google webmaster center</a>
```

plugins/Sitemaps/class.sitemaps.plugin.php
```diff
@@ -1,4 +1,7 @@
-<?php if (!defined('APPLICATION')) { exit; }
+<?php
+use Vanilla\Contracts\ConfigurationInterface;
+
+if (!defined('APPLICATION')) { exit; }
```
initvector: It's probably safe to nuke this line. There is only a class definition in here.

PR #715, opened 2019-06-06
Fix: category discussion type setting not respected when creating a question

dannimajdalani:
related [#426](https://github.com/vanilla/support/issues/426)

### Issue

When a category has custom permissions that does not allow questions, users can still for post a question in the category by browsing to post/question/no-question-category

### Case QnA

When a new question is created, [the QnA addon's postController_question_create hook](https://github.com/vanilla/addons/blob/d483877ea17a8578f891be621a71e15b4783bc16/plugins/QnA/class.qna.plugin.php#L1115) is called that calls [`PostController::discussion`](https://github.com/vanilla/vanilla/blob/4df04723ba501b050951ef1c4afd51d92d072391/applications/vanilla/controllers/class.postcontroller.php#L112). We're not checking the allowed discussion types when creating the discussion. 

### Proposed Fix - To stop force posting
In [the QnA addon's postController_question_create hook](https://github.com/vanilla/addons/blob/d483877ea17a8578f891be621a71e15b4783bc16/plugins/QnA/class.qna.plugin.php#L1115) we can get the category and check if the allowed discussion type is a question, since we are in the QnA plugin. Otherwise, we can throw an exception.

```
        $categoryModel = new CategoryModel();
        $category = (array)$categoryModel->getByCode($categoryUrlCode);
        $category = $categoryModel::permissionCategory($category);
        $isAllowedTypes = isset($category['AllowedDiscussionTypes']);
        $isAllowedQuestion = in_array('Question', $category['allowedDiscussionType']);
        if ($category &&  !$isAllowedQuestion && $isAllowedTypes) {
            $sender->Form->addError(t('You are not allowed to post a question in this category.'));
        }
```
### DiscussionType

in [QnA/views/post.php#L24](https://github.com/vanilla/addons/blob/master/plugins/QnA/views/post.php#L24) we are using   'PermFilter' => ['AllowedDiscussionTypes' => 'Question'], after that we call categoryDropDown('CategoryID', $options); categoryDropDown then tries to get the discussiontypes in [class.form.php#L500](https://github.com/vanilla/vanilla/blob/master/library/core/class.form.php#L500) which return nothing since we sent `'PermFilter' => ['AllowedDiscussionTypes' => 'Question']` instead of `'DiscussionType'  => 'Question']`
### Case Polls

Similar to QnA, we can check for allowed discussion types in [the Polls addon's postController_poll_create hook](https://github.com/vanilla/internal/blob/d42027f7b53f25e07459afeb5e4a7fe419d96766/plugins/Polls/class.polls.plugin.php#L275).

Screen Shot 2019-05-31 at 11 22 56 AM

### To Test

- [ ] Create a Category that is only restricted to (Discussions) - i'll call it no-questions
- [ ] from the dropdown on the main page, click new question (you should only see categories that allow questions)
- [ ] try again, this time go to post/question/no-question (you should get an error)
- [ ] Repeat the above using sub-communties

### DIFF
--- a/plugins/QnA/class.qna.plugin.php
+++ b/plugins/QnA/class.qna.plugin.php
@@ -1113,14 +1113,12 @@ class QnAPlugin extends Gdn_Plugin {
      * @param PostController $sender Sending controller instance.
      */
     public function postController_question_create($sender, $categoryUrlCode = '') {
-        $category = false;
-        if ($categoryUrlCode != '') {
-            $categoryModel = new CategoryModel();
-            $category = (array)$categoryModel->getByCode($categoryUrlCode);
-            $category = $categoryModel::permissionCategory($category);
-            $isAllowedTypes = isset($category['AllowedDiscussionTypes']);
-            $isAllowedQuestion = in_array('Question', $category['allowedDiscussionType']);
-        }
+        $categoryModel = new CategoryModel();
+        $category = (array)$categoryModel->getByCode($categoryUrlCode);
+        $category = $categoryModel::permissionCategory($category);
+        $isAllowedTypes = isset($category['AllowedDiscussionTypes']);
+        $isAllowedQuestion = in_array('Question', $category['allowedDiscussionType']);
+
         if ($category &&  !$isAllowedQuestion && $isAllowedTypes) {
             $sender->Form->addError(t('You are not allowed to post a question in this category.'));
         }
--- a/plugins/QnA/class.qna.plugin.php
+++ b/plugins/QnA/class.qna.plugin.php
@@ -1113,16 +1113,16 @@ class QnAPlugin extends Gdn_Plugin {
      * @param PostController $sender Sending controller instance.
      */
     public function postController_question_create($sender, $categoryUrlCode = '') {
-        $requestCategoryID = GDN::Request()->post('CategoryID');
-        $category = CategoryModel::categories($requestCategoryID);
-
-        if ($category && isset($category['AllowedDiscussionTypes'])) {
-            foreach($category['AllowedDiscussionTypes'] as $allowedDiscussionType) {
-                $allowedType[] = $allowedDiscussionType;
-            }
-            if (!in_array('Question', $allowedType)) {
-                $sender->Form->addError('You are not allowed to post a question in this category.');
-            }
+        $category = false;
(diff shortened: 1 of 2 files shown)


plugins/QnA/class.qna.plugin.php
```diff
@@ -1113,10 +1113,21 @@ public function postController_afterForms_handler($sender) {
...
     public function postController_question_create($sender, $categoryUrlCode = '') {
-            // Create & call PostController->discussion()
-            $sender->View = PATH_PLUGINS . '/QnA/views/post.php';
-            $sender->setData('Type', 'Question');
-            $sender->discussion($categoryUrlCode);
+        $categoryModel = new CategoryModel();
+        if ($categoryUrlCode != '') {
+            $category = (array)$categoryModel->getByCode($categoryUrlCode);
+            $category = $categoryModel::permissionCategory($category);
+            $isAllowedTypes = isset($category['AllowedDiscussionTypes']);
```
cchabilall83: maybe we should be checking is the array_key_exists?

plugins/QnA/class.qna.plugin.php
```diff
@@ -1113,10 +1113,21 @@ public function postController_afterForms_handler($sender) {
...
+            $isAllowedQuestion = in_array('Question', $category['allowedDiscussionType']);
+        }
```
cchabilall83: I think you might have casing error here.  
$category['allowedDiscussionType'] vs $category['AllowedDiscussionTypes']

plugins/QnA/class.qna.plugin.php
```diff
@@ -1113,10 +1113,21 @@ public function postController_afterForms_handler($sender) {
...
+
+        if ($category && !$isAllowedQuestion && $isAllowedTypes) {
+            $sender->Form->addError(t('You are not allowed to post a question in this category.'));
```
cchabilall83: Shouldn't we just have to check if AllowedDiscussionTypes has type 'Question' in it?  $isAllowedTypes I don't believe is necessary

plugins/QnA/class.qna.plugin.php
dannimajdalani: Good catch. It was getting set to false and because of the wrong key

PR #726, opened 2019-07-16
Fix database name validation in Forum Merge addon

alex-mtl:
Validate form value 'Database',
other form values sanitized through PDO->prepare when query

Closes: https://github.com/vanilla/vanilla-patches/issues/423

### DIFF
--- a/plugins/ForumMerge/class.forummerge.plugin.php
+++ b/plugins/ForumMerge/class.forummerge.plugin.php
@@ -47,7 +47,7 @@ class ForumMergePlugin implements Gdn_IPlugin {
      * @return bool
      */
    private function validateForm(Gdn_Form &$form): bool {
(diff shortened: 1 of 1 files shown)


plugins/ForumMerge/class.forummerge.plugin.php
```diff
@@ -28,16 +28,37 @@ public function utilityController_merge_create($sender) {
...
    }
 
+    /**
+     * Validate form values.
+     *
+     * @param Gdn_Form $form
+     * @return bool
+     */
+   private function validateForm(Gdn_Form &$form): bool {
+       //Validate form value 'Database', other form values sanitized through PDO->prepare when query
```
charrondev:
```suggestion
       // Validate form value 'Database', other form values sanitized through PDO->prepare when query
```

PR #727, opened 2019-07-23
Authenticate the postback in the no index addon

tburry:
The no index addon was not checking for an authenticated post back which is a narrow CSRF vulnerability. This PR also cleans up the spacing in the method and updates the parameter to the current best practice.

Closes https://github.com/vanilla/vanilla-patches/issues/268.


tburry:
Just a n FYI: I fixed some old notices in this plugin after this PR, but that PR got merged first. I'm not sure if there will be any weirdness because of that.

PR #728, opened 2019-07-26
Update CloudflareSupport

kaecyra:


tburry:
@kaecyra who do you expect would review this? Can we just merge and see if the OSS community finds it useful?

kaecyra:
Sure.

PR #731, opened 2019-08-08
Update the site maps addon to point to discussions directly

tburry:
Adds a feature flag to render site maps that point to discussions directly rather than category archives (see SitemapsPlugin::DISCUSSION_SITE_MAPS) for the feature name.

The new sitemaps use the index to point to categories that page discussions by a limit of 1000 which is a number we may tweak. This ends up putting more entries in the site map index too (as much as ~6000 for some sites), but the hope is since the sitemap index is largely generated programmatically we should be okay with performance.

The reason for this change is:

1. We get several complaints about the category archive pages not allowing indexing because that shows as an error in Google Webmaster tools after a more recent change.
2. There is an edge case where we may want to disallow access to most of the site except discussions for certain SEO trickery.

This change is being put behind a feature flag because it impacts SEO and whenever we make a change to SEO related functionality we don’t know whether the change will be good, bad, or neutral so we have to take care and roll it out slowly.


tburry:
I wonder if you need to merge master into this to get the build working?

PR #734, opened 2019-08-20
Remove the Transient Key from the URL in the Language Selector.

nervoustwit:
We have links to change a user's language preference by GET request that pass the Transient Key in the address. This can be considered a security vulnerability because it exposes the Transient Key in logs, etc.

This PR adds the `.js-hijack` class to the link to change the language so that the flyout.js will then turn that GET request into a POST request and pass the Transient Key with it.

### Test

- Turn on the Multilingual Plugin, configure a few languages in the Locales section.
- Look at the URL of the Footer.
- See that the TransientKey is one of the parameters.
- Checkout this branch.
- See that the Transient Key is no longer there.
- Click on the link and make sure that the language is changed.

### DIFF
--- a/plugins/Multilingual/class.multilingual.plugin.php
+++ b/plugins/Multilingual/class.multilingual.plugin.php
@@ -128,13 +128,14 @@ class MultilingualPlugin extends Gdn_Plugin {
     /**
      * Allow user to set their preferred locale via link-click.
      */
-    public function profileController_setLocale_create($sender, $locale, $tK) {
+    public function profileController_setLocale_create($sender, $locale) {
         if (!Gdn::session()->UserID) {
             throw permissionException('Garden.SignIn.Allow');
         }
 
+        $tk = gdn::request()->post('TransientKey');
         // Check intent.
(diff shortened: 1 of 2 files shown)


nervoustwit:
### BEFORE MERGING
I am doing a review of our custom Themes to make sure that this does not affect them.

plugins/Multilingual/class.multilingual.plugin.php
```diff
@@ -128,13 +128,14 @@ protected function getAlternateLocale() {
...
-    public function profileController_setLocale_create($sender, $locale, $tK) {
+    public function profileController_setLocale_create($sender, $locale) {
         if (!Gdn::session()->UserID) {
             throw permissionException('Garden.SignIn.Allow');
         }
 
+        $tk = gdn::request()->post('TransientKey');
         // Check intent.
-        if (!Gdn::session()->validateTransientKey($tK)) {
+        if (!Gdn::session()->validateTransientKey($tk)) {
```
initvector: Could you check for "authenticated postback" here, instead of manually retrieving and validating the transient key?

PR #763, opened 2019-11-21
[QnA]: Fix checking of curation manage permission when accepting/rejecting a…

dannimajdalani:
closes https://github.com/vanilla/support/issues/656

### Details

Users with Garden.Curation.Manage can accept answers, however, they cannot reject them or set them to unasnwered because of https://github.com/vanilla/addons/blob/master/plugins/QnA/class.qna.plugin.php#L267

We are checking if users have Vanilla.Discussions.Edit

### To Test

As a user, create a question, add an answer
With another user that has curation manage permission try to edit the question

### DIFF
--- a/plugins/QnA/class.qna.plugin.php
+++ b/plugins/QnA/class.qna.plugin.php
@@ -264,10 +264,10 @@ class QnAPlugin extends Gdn_Plugin {
             return;
         }
 
-        if (!Gdn::session()->checkRankedPermission('Garden.Curation.Manage')) {
-            if (!Gdn::session()->checkPermission('Vanilla.Discussions.Edit', true, 'Category', $discussion->PermissionCategoryID)) {
-                return;
-            }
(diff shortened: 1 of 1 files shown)


plugins/QnA/class.qna.plugin.php
```diff
@@ -264,10 +264,11 @@ public function discussionController_commentOptions_handler($sender, $args) {
             return;
         }
 
-        if (!Gdn::session()->checkPermission('Vanilla.Discussions.Edit', true, 'Category', $discussion->PermissionCategoryID)) {
-            return;
+        if (!Gdn::session()->checkRankedPermission('Garden.Curation.Manage')) {
```
tburry: Can you just do this as an `||` rather than a nested `if`?

dannimajdalani: !(A || B) = !A && !B much better than a nested if.

PR #766, opened 2019-12-17
Add $pager parameter to getUnansweredCount function in Q&A plugin.

patches:
There is currently no way to determine if the unanswered count is affecting the Pager when hooking into the `unansweredCount` event.

The $pager parameter has been added to the `getUnansweredCount` function and is passed as an argument to the `unansweredCount` event. False is the default value assigned to this variable.

### DIFF
--- a/plugins/QnA/class.qna.plugin.php
+++ b/plugins/QnA/class.qna.plugin.php
@@ -925,7 +925,7 @@ class QnAPlugin extends Gdn_Plugin {
      */
     public function discussionsController_beforeBuildPager_handler($sender, $args) {
         if (Gdn::controller()->RequestMethod == 'unanswered') {
-            $count = $this->getUnansweredCount();
+            $count = $this->getUnansweredCount(true);
             $sender->setData('CountDiscussions', $count);
         }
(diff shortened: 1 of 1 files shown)


plugins/QnA/class.qna.plugin.php
```diff
@@ -925,17 +925,18 @@ public function discussionsController_unanswered_create($sender, $args) {
...
      * Return the number of unanswered questions.
      *
+     * @param bool $pager Whether this count affects the pager or not.
      * @return int
      */
-    public function getUnansweredCount() {
+    public function getUnansweredCount($pager = false) {
         // TODO: Dekludge this when category permissions are refactored (tburry).
         $cacheKey = Gdn::request()->webRoot().'/QnA-UnansweredCount';
         $questionCount = Gdn::cache()->get($cacheKey);
```
alex-mtl: If we add extra parameter here, we should either differ `$cacheKey` for 2 different values or bypass caching for one of them.

I would prefer probably to go with 1st attempt. Implement 2 `$cacheKey` variants.

PR #771, opened 2020-02-06
Add no-cache header to jsconnect

igraziatto:
Closes https://github.com/vanilla/support/issues/1312

The issue seems to be caused because `entry/jsconnect` was being cached for around 2 minutes, so when then client SSO requested it was returning a nonce that no longer existed. (Nonce is deleted as soon as it's used).

I guess the way of testing this is verifying if `entry/jsconnect` is returning `Cache-Control: no-cache` header. You can try setting up https://github.com/vanilla/stub-sso-providers too and testing against it too.


nervoustwit:
@initvector Just checking. It would be better to set headers using the CacheControlMiddleware?

`\Vanilla\Web\CacheControlMiddleware::sendCacheControlHeaders(\Vanilla\Web\CacheControlMiddleware::NO_CACHE);`

Issue #779, opened 2020-03-05
Forums redirector plugin doesn't handle lithium category redirect properly. 


vincentlam92:
Issue reference: https://github.com/vanillaforums/F-Secure/issues/1#issuecomment-594718974

## For lithium, their Board(bd-p) + Category(ct-p) = Vanilla Category. But our redirector only handles board redirect.  And also their KB categories have 'tkb-p' which is not supported as well. 
```
Example Url: https://community.clientname.com/t5/catgeoryName/**ct-p**/urlcode  -> (not handle by redirector )
```
---
```
Example Url: https://community.clientname.com/t5/catgeoryName/**tkb-p**/urlcode  -> (not handle by redirector )
```
---
```
Example Url: https://community.clientname.com/t5/catgeoryName/**bd-p**/urlcode -> (redirector handle since is from board)
```

### temperary solution 
- create Route :
```t5\/.*?\/ct-p\/(.*)$``` -> categories/$1
also
```t5\/.*?\/tkb-p\/(.*)$``` -> categories/$1

vincentlam92:
pr:https://github.com/vanilla/addons/pull/782

Issue #737, opened 2019-08-27
Forum Redirector: PunBB redirects not properly handled when there is a number in discussion names.


vincentlam92:
The redirector plugin does not redirect to the correct url when a discussion has a number in its name.

 **For example:**
```
domain.com/topic/1/test-number-3212
domain.com/topic/100/20th-year-anniversary
```
 **will be redirected to** 
```
domain.com/discussion/1/test-number-3212/p3212
domain.com/discussion/100/20th-year-anniversary/p20
```
 **Redirects will work properly otherwise.**

vincentlam92:
already submit the code ,but didnt create a pr yet as per pat told

vincentlam92:
The pr link :https://github.com/vanilla/addons/pull/740

PR #782, opened 2020-03-13
Fix redirector plugin not properly handling redirect of Lithium 'ct-p' and 'kb-t' subfolders.

vincentlam92:
## punbb 
issue :https://github.com/vanilla/addons/issues/737
The redirector plugin does not redirect to the correct url when a discussion has a number in its name. It will take that number and become a page number.

In the end, this issue still reminds pending since we need more investigation.  

----- 
## lithium 
issue:https://github.com/vanilla/addons/issues/779

The redirector plugin does not handle the category type **('ct-p')** and **tkb-p**,  I added a or condition while we get the Url from Lithium

### DIFF
--- a/plugins/Redirector/class.redirector.plugin.php
+++ b/plugins/Redirector/class.redirector.plugin.php
@@ -385,7 +385,7 @@ class RedirectorPlugin extends Gdn_Plugin {
             if (val('_arg3', $get) == 'page') {
                 $result['_arg4'] = 'Page';
             }
-        } elseif (val('_arg1', $get) == 'bd-p') { // Board = Category
+        } elseif (val('_arg1', $get) == 'bd-p' || $get['_arg1'] == 'ct-p') { // bd-p ,ct-p ,tkb-p are Category, but pending for tkb-p
             $result = [
                 '_arg2' => [
                     'CategoryCode',
@@ -540,7 +540,7 @@ class RedirectorPlugin extends Gdn_Plugin {
             ];
             self::vbFriendlyUrlID($get, 't');
         }
-        
+
         return $data;
 
     }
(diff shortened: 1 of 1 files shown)


plugins/Redirector/class.redirector.plugin.php
```diff
@@ -602,18 +602,24 @@ public static function topicFilter(&$get) {
...
-                '_arg0' => [
-                    'DiscussionID',
-                    'Filter' => [__CLASS__, 'removeID'],
-                ],
-                '_arg1' => [
-                    'Page',
-                    'Filter' => [__CLASS__, 'IPBPageNumber'],
-                ],
-            ];
+            if (!isset($get['_arg3'])) {
```
patches: I understand that this might work for PunBB.

My main concern here was mostly for IPB migrations. You base your check simply on the fact that the 3rd parameter is not set to determine if it's either a PunBB or a IPB URL.

The example you gave in the last PR has a page number which means the 3rd parameter will be set and in this case the page number will be retrieved from the Url.

However, what happens if there is no page number in the Url as in the example below?

topic/403777-40-permalinkurl-structure-options/

This will be considered as a PunBB url when in fact, it's an IPB url and `403777-40-permalinkurl-structure-options` will be returned as the discussion id.

Here is the result you will get as an example.

Screen Shot 2020-03-16 at 3 14 20 PM

plugins/Redirector/class.redirector.plugin.php
```diff
@@ -385,7 +385,7 @@ public static function t5Filter(&$get) {
             if (val('_arg3', $get) == 'page') {
                 $result['_arg4'] = 'Page';
             }
-        } elseif (val('_arg1', $get) == 'bd-p') { // Board = Category
+        } elseif (val('_arg1', $get) == 'bd-p' || $get['_arg1'] == 'ct-p') { // bd-p ,ct-p ,tkb-p are Category, but pending for tkb-p
```
patches: What do you think of something like this instead.

Also, never put the comment at the end of a line. As per our standards, a line cannot exceed 80 characters. Always put the comment on the line above. It also improves readability.

```suggestion
        } elseif (isset($get['_arg1']) && in_array($get['_arg1'], ['bd-p', 'ct-p'])) {
```

vincentlam92: Actually, in here, I have a question for KB. Lithium kb type of DIscussion will have ```www.urlxxxx/tkb-p/xxxxx.com``. Should I add that in the in_array as well? Or we will handle Kb redirect differently?

plugins/Redirector/class.redirector.plugin.php
vincentlam92: Oh I see, sorry I never actually see how IPB Url looks like for discussion, and lack of consider for IPB. Thank to point it out.

vincentlam92: @patches , and also another thing i will like to confirm. Do you think the example I found will be the standard PunBB discussion URL with page number? 

punbb example with page number:
```
http://punbb.informer.com/forums/topic/10778/what-antivirus-do-you-use/page/2/
```
If it is , the page number should be at arg position 5? like **'_arg5'**?

PR #798, opened 2020-04-14
Fix discussion API failure when no question status is set

dannimajdalani:
closes https://github.com/vanilla/support/issues/1699

### Details

Making an api/v2/discussions call When a Discussion of Type question and QnA Null would fail with this error
```
"message": "Validation Failed",

"status": 422,

"errors": [

    {

        "field": "status",

        "code": "missingField",

        "path": "[22].attributes.question",

        "status": 422,

        "message": "item[22].attributes.question.status is required."

    }

]
```

This is happening because of https://github.com/vanilla/addons/blob/master/plugins/QnA/class.qna.plugin.php#L1320
w'ere not setting a default value when that field is null.

### DIFF
--- a/plugins/QnA/class.qna.plugin.php
+++ b/plugins/QnA/class.qna.plugin.php
@@ -1317,7 +1317,7 @@ class QnAPlugin extends Gdn_Plugin {
         }
 
         $discussion['attributes']['question'] = [
-            'status' => strtolower($discussion['qnA']),
+            'status' => isset($discussion['qnA']) ? strtolower($discussion['qnA']) : 'unanswered',
             'dateAccepted' => $discussion['dateAccepted'],
             'dateAnswered' => $discussion['dateOfAnswer'],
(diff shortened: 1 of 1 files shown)


plugins/QnA/class.qna.plugin.php
```diff
@@ -1317,7 +1317,7 @@ public function discussionsApiController_normalizeOutput(
         }
 
         $discussion['attributes']['question'] = [
-            'status' => strtolower($discussion['qnA']),
+            'status' => isset($discussion['qnA']) ? strtolower($discussion['qnA']) : 'unanswered',
```
tburry, line 1320:
```suggestion
            'status' => empty($discussion['qnA']) ? strtolower($discussion['qnA']) : 'unanswered',
```

I suggest `empty` over `isset` because it handles one more potential data error. I just fixed a different bug with this exact case so I may be a bit more paranoid.

PR #801, opened 2020-04-17
Fix Subcommunity conflict on Redirect URL

rbergDrox:
Issue on **Category As Link** plugin : root directory twice in the url.

### DIFF
--- a/plugins/categoryaslink/class.categoryaslink.plugin.php
+++ b/plugins/categoryaslink/class.categoryaslink.plugin.php
@@ -120,10 +120,9 @@ if (!function_exists("categoryUrl")) {
         if (class_exists('SubcommunitiesPlugin') && $category['RedirectUrl']) {
             $requestRoot = Gdn::request()->getRoot();
             // Remove $requestRoot in $destination because it will appear twice because of url() or safeURL().
-            $destination = $requestRoot && strpos($category['RedirectUrl'], $requestRoot) === 0 ?
-                substr($category['RedirectUrl'], strlen($requestRoot)) : $category['RedirectUrl'];
-
-            return safeURL($destination);
+            $destination = safeURL($category['RedirectUrl']);
+            return $requestRoot && strpos($destination, $requestRoot) === 0 ?
+                substr($destination, strlen($requestRoot)) : $destination;
         }
 
         // If there is a URL that links to a discussion it overrides the category or even a link to an alias category.
--- a/plugins/categoryaslink/class.categoryaslink.plugin.php
+++ b/plugins/categoryaslink/class.categoryaslink.plugin.php
@@ -91,9 +91,14 @@ class CategoryAsLinkPlugin extends Gdn_Plugin {
      * @param CategoriesController $sender
      * @param CategoriesController $args
      */
-    public function categoriesController_beforeCategoriesRender_handler($sender, $args) {
+    public function categoriesController_beforeCategoriesRender_handler(CategoriesController $sender, $args) {
         if ($sender->data('Category.RedirectUrl')) {
-            redirectTo(categoryUrl($sender->Data('Category')), 301);
+            $destination = categoryUrl($sender->Data('Category'));
+            $requestRoot = Gdn::request()->getRoot();
+            if ($requestRoot && strpos($destination, $requestRoot) === 0) {
+                $destination = substr($destination, strlen($requestRoot));
+            }
+            redirectTo($destination, 301);
         }
     }
 }
--- a/plugins/categoryaslink/class.categoryaslink.plugin.php
+++ b/plugins/categoryaslink/class.categoryaslink.plugin.php
@@ -95,6 +95,8 @@ class CategoryAsLinkPlugin extends Gdn_Plugin {
         if ($sender->data('Category.RedirectUrl')) {
             $destination = categoryUrl($sender->Data('Category'));
             $requestRoot = Gdn::request()->getRoot();
+            // Remove $requestRoot in $destination because it will appear twice because of url() or safeURL()
+            // usage from redirectTo().
             if ($requestRoot && strpos($destination, $requestRoot) === 0) {
                 $destination = substr($destination, strlen($requestRoot));
             }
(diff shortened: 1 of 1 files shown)


plugins/categoryaslink/class.categoryaslink.plugin.php
```diff
@@ -91,9 +91,16 @@ public function categoriesController_render_before($sender, $args) {
...
         if ($sender->data('Category.RedirectUrl')) {
-            redirectTo(categoryUrl($sender->Data('Category')), 301);
+            $destination = categoryUrl($sender->Data('Category'));
+            $requestRoot = Gdn::request()->getRoot();
+            // Remove $requestRoot in $destination because it will appear twice because of url() or safeURL()
+            // usage from redirectTo().
+            if ($requestRoot && strpos($destination, $requestRoot) === 0) {
+                $destination = substr($destination, strlen($requestRoot));
+            }
+            redirectTo($destination, 301);
```
igraziatto, line 103: I don't think the issue is here but in the `categoryUrl()` function override just below.

The issue is that if there's a `RedirectUrl` it returns the value of `RedirectUrl` wrapped in `safeURL()` without considering Subcommunities.
Then it follows to consider Subcommunities, but only below, around line 126.

My suggestion is to leave `categoriesController_beforeCategoriesRender_handler` as is and modify `categoryUrl()` slightly to check if `class_exists('SubcommunitiesPlugin') && $category['RedirectUrl']` and then apply your logic of removing the root.

rbergDrox: @igraziatto 
good call. I moved my logic

plugins/categoryaslink/class.categoryaslink.plugin.php
```diff
@@ -116,6 +116,15 @@ function categoryUrl($category, $page = '', $withDomain = true) {
...
         $category = (array)$category;
 
+        // If SubcommunitiesPlugin is enabled and the category use RedirectUrl feature.
+        if (class_exists('SubcommunitiesPlugin') && $category['RedirectUrl']) {
+            $requestRoot = Gdn::request()->getRoot();
+            // Remove $requestRoot in $destination because it will appear twice because of url() or safeURL().
+            $destination = safeURL($category['RedirectUrl']);
+            return $requestRoot && strpos($destination, $requestRoot) === 0 ?
+                substr($destination, strlen($requestRoot)) : $destination;
+        }
```
igraziatto, line 126: This code works, there's nothing wrong with it per se, but you're checking for `$category['RedirectUrl']` twice. If you checked for `$category['RedirectUrl']` first and then inside that block check for `class_exists('SubcommunitiesPlugin')` and handle both scenarios with an `else` statement this baby would really shine.

plugins/categoryaslink/class.categoryaslink.plugin.php
```diff
@@ -118,8 +118,19 @@ function categoryUrl($category, $page = '', $withDomain = true) {
...
+                // Remove $requestRoot in $destination because it will appear twice because of url() or safeURL().
+                $destination = safeURL($category['RedirectUrl']);
+                $redirectUrlDestination = $requestRoot && strpos($destination, $requestRoot) === 0 ?
+                    substr($destination, strlen($requestRoot)) : $destination;
+            } else {
+                // SafeURL because you may be linking to another web property, another forum or knowledgebase.
+                $redirectUrlDestination = safeURL(val('RedirectUrl', $category));
+            }
+
+            return $redirectUrlDestination;
```
charrondev, line 133: This is not really a sufficient solution.

Subcommunities is not the only thing we have that can modify the base community URL. (hub/node, reverseproxy plugin).

Additionally this is not the only plugin using `safeURL()`. Maybe `safeURL` is what should be updated here.

No core plugin should **_ever_** be doing `if (SubcommunityPlugin)`. It's a big hack, and isn't particularly robust.

Slack me if you want help finding a better solution.

rbergDrox: @charrondev I used `if (class_exists('SubcommunitiesPlugin')) {` because of existing  code with the condition.  `safeURL`  is used twice because of the redirectionTo in`categoriesController_beforeCategoriesRender_handler`. I was thinking of historical reasons  it was  like that. In fact we can return `$category['RedirectUrl']` directly

charrondev: That works. Do you know if that condition was in the product or in the services repos?

This type of plugin checking is generally verboten on the product side, as it leads to very fragile code. You can sometimes get away with it in a singular customer repo because the customer has a much narrower set of site configuration possbilities.

rbergDrox: You mean this kind of code? 
```php      
if (class_exists('SubcommunitiesPlugin')) {
            // Subcommunities version.
            $path = '/categories/'.rawurlencode(val('UrlCode', $category));
            $categoryURL = Gdn::getContainer()
                ->get(SubcommunitiesPlugin::class)
                ->subcommunityURL($category['categoryID'] ?? 0, $path, $withDomain, $page, SubcommunitiesPlugin::URL_TYPE_CATEGORY);
        } elseif (class_exists('SEOLinksPlugin')) {
```

igraziatto: @charrondev `class_exists('SubcommunitiesPlugin')` was already in the code. **Categories as Links** is an old plugin.

This PR is related to [this issue](https://github.com/vanilla/support/issues/1726) which was escalated to Services because it started presenting issues after Asus started using a custom [redirector addon](https://github.com/vanillaforums/asus/tree/master/plugins/zentalkredirector) but since it seems it's core-related maybe it shouldn't be up to Services to fix. Also seems more complex than a tweak in `categoryaslink.plugin.php` as you mentioned so maybe we should close this PR and start over. ICB can take the issue back.

charrondev: The thing is `CategoryAsLink` isn't that old of a plugin. It's only from the last year and a half.

In any case that sounds like a good idea.

PR #802, opened 2020-04-20
Fix Cloudflare plugin’s loop continue

tburry:
This is currently a warning in our PHP 7.3 build so it would be nice to fix it.

### DIFF
--- a/plugins/CloudflareSupport/class.cloudflaresupport.plugin.php
+++ b/plugins/CloudflareSupport/class.cloudflaresupport.plugin.php
@@ -62,14 +62,12 @@ class CloudflareSupportPlugin extends Gdn_Plugin {
 
         // Iterate and see if connecting IP is from Cloudflare
         foreach ($currentRanges as $cloudflareIPRange) {
-
-            // Check if calling IP is a Cloudflare proxy
+            // Check if calling IP is a Cloudflare proxy.
             switch ($ipType) {
-
                 // IPv4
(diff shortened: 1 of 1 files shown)


plugins/CloudflareSupport/class.cloudflaresupport.plugin.php
```diff
@@ -62,14 +62,12 @@ public function __construct() {
...
-
-            // Check if calling IP is a Cloudflare proxy
+            // Check if calling IP is a Cloudflare proxy.
             switch ($ipType) {
-
                 // IPv4
                 case self::CF_IPV4:
                     if (!$this->ipv4_in_range($requestAddress, $cloudflareIPRange)) {
-                        continue;
+                        continue 2;
```
tburry, line 70: The specific warning I get in PHP 7.3 is that `continue` targets the switch and not the loop.

initvector: Sounds good!

I don't think we see a ton of switch cases in loops, so in case anyone following is interested:

>  In PHP the switch statement is considered a looping structure for the purposes of continue. continue behaves like break (when no arguments are passed) but will raise a warning as this is likely to be a mistake. If a switch is inside a loop, continue 2 will continue with the next iteration of the outer loop.

[PHP: continue](https://www.php.net/manual/en/control-structures.continue.php)

PR #804, opened 2020-05-04
Fix BBCode format when IPB Fomatter is enabled

igraziatto:
Closes https://github.com/vanilla/support/issues/1562

**Steps to test:**
- Enable the "IPB Formatter" plugin (addons)
- Set the posting format to "BBCode"
- Make a comment and using the editor emoji menu, create a heart emoji
- Put in a line break (press return/enter), make another heart emoji
- Create comment
- the `<br>` line break tag should be escaped (bug)
- Checkout this branch

**Further testing:**
- Make another BBCode comment and add a some BBCode, maybe some acceptable HTML
- Change the format in the DB to "ipb"
- Check if the comment renders alright.

### DIFF
--- a/plugins/IPBFormatter/class.ipbformatter.plugin.php
+++ b/plugins/IPBFormatter/class.ipbformatter.plugin.php
@@ -29,9 +29,9 @@ class IPBFormatterPlugin extends Gdn_Plugin {
         $formatService = $dic->get(FormatService::class);
         $formatService->registerFormat(IPBFormat::FORMAT_KEY, $dic->get(IPBFormat::class));
 
-        $dic->rule("IPBFormatter")
-            ->setClass(Formatter::class)
-            ->setShared(true);
+//        $dic->rule("IPBFormatter")
+//            ->setClass(Formatter::class)
+//            ->setShared(true);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


plugins/IPBFormatter/class.ipbformatter.plugin.php
```diff
@@ -29,9 +29,9 @@ public function container_init(ContainerInterface $dic) {
         $formatService = $dic->get(FormatService::class);
         $formatService->registerFormat(IPBFormat::FORMAT_KEY, $dic->get(IPBFormat::class));
 
-        $dic->rule("IPBFormatter")
-            ->setClass(Formatter::class)
-            ->setShared(true);
+//        $dic->rule("IPBFormatter")
+//            ->setClass(Formatter::class)
+//            ->setShared(true);
```
charrondev, line 34: Commented lines should be removed.

plugins/IPBFormatter/class.ipbformatter.plugin.php
```diff
@@ -29,9 +29,9 @@ public function container_init(ContainerInterface $dic) {
         $formatService = $dic->get(FormatService::class);
         $formatService->registerFormat(IPBFormat::FORMAT_KEY, $dic->get(IPBFormat::class));
```
charrondev, line 30: I'd recommend moving these to the bootstrap file as well. `container_init` could cause some race conditions if something obtains a copy of the format service during plugin initialization.

Moving it to bootstrap ensures this race condition doesn't exist.

tburry: You want to configure the container, not use it. You want something like this:

```php
$dic->rule(FormatService::class)
    ->addCall('registerFormat', [IPBFormat::FORMAT_KEY, new Reference(IPBFormat::class)])
```

This configures the container lazily so that it registers the IPB format when the format service is created.

PR #805, opened 2020-05-11
Fix admins able to spoof system user

blutarch:
Closes vanilla/support#1869

An admin user is able to spoof in as the system user. This is undesirable because they can then see themes and plugins we may not want them to. This adds a check before displaying the spoof button and prevents it from being displayed for system users.

### TO TEST
1. As an admin, go to `dashboard/user`
1. Verify that no system users have spoof buttons.


dannimajdalani:
You'll need to assign someone for review @blutarch

PR #806, opened 2020-05-22
Filter meta tags with Civil Tongue

blutarch:
Closes vanilla/support#1461

This PR fixes a problem where meta tags aren't being changed by Civil Tongue, resulting in the bad words reappearing when posts are shared via twitter or facebook. This PR replaces appearances of the bad words in meta tags with the chosen replacement.

### TO TEST
1. Enable Civil Tongue and set up a forbidden word and its replacement.
1. Make a post with the forbidden word.
1. Using dev tools, look at the meta tags and note that the forbidden words appear in the `content` attribute of meta tags.
1. Check out this branch and verify that the words have been replaced.

### DIFF
--- a/plugins/CivilTongueEx/class.civiltongueex.plugin.php
+++ b/plugins/CivilTongueEx/class.civiltongueex.plugin.php
@@ -566,6 +566,23 @@ class CivilTonguePlugin extends Gdn_Plugin {
     }
 
     /**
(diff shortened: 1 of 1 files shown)


nervoustwit:
I have tried to test this locally, it appears to work but in order to be sure we would have to deploy this to a live site with a legit Facebook (I didn't see Twitter scraping my message and putting it in the Tweet) integration.

PR #807, opened 2020-05-28
Fix JSConnect v3 setting the target

igraziatto:
Closes https://github.com/vanilla/support/issues/1933

**Steps to test:**
1. Make sure your local is set to work with jsConnect. You can test with the sso stub: https://github.com/vanilla/stub-sso-providers
2. Enable jsConnect and configure the addon following the stub guidelines.
3. Set jsConnect protocol to v3
4. Add the param `redirect={target}` to "Sign in URL". This param should send the user back to the page they were after connecting.
5. In an anonymous tab, from another page that it's **not** the homepage (a discussion for example) log in through jsConnect.
6. Observe you are redirected to the homepage after connecting.
7. Checkout this branch repeat step 5. This time it should redirect to the page you were before.

### DIFF
--- a/plugins/jsconnect/class.jsconnect.plugin.php
+++ b/plugins/jsconnect/class.jsconnect.plugin.php
@@ -531,7 +531,7 @@ class JsConnectPlugin extends SSOAddon {
             if ($protocol !== self::PROTOCOL_V3) {
                 $this->entryJsConnectV2($sender, $action, $target);
             } else {
-                $this->entryJsConnectV3($sender);
+                $this->entryJsConnectV3($sender, $target);
             }
         } else {
             $sender->addDefinition('jsconnect', [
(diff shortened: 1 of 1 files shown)


nervoustwit:
When you don’t have JSConnect set as the default sign in method, it loses the Target. You click on Sign In, you go to our Sign In page and then when you mouse over the JSConnect Sign In button you see already the target is set to “/”.

The problem is in the method that creates the button:

https://github.com/vanilla/addons/blob/2a3e6d60aee19647eca575cd24257ed8c8189e47/plugins/jsconnect/class.jsconnect.plugin.php#L176 

It looks for `target` in the GET and if it doesn't find it, it looks for `target` in the GET again. It should look for `Target` the second time (or the first).

Issue #816, opened 2020-06-10
Add Q&A status badge to the discussion


tburry:
Currently, when viewing a question, you can see the Q&A status of each comment, but you can't see the Q&A status of the question (discussion) itself.

Let's add the status to the OP and have it user the same look/feel as the discussion list. It can be after the date like the comments.

I think this may also need a feature flag as existing customers may have had custom theming to add it redundantly (we have specific feature flag functionality).

PR #818, opened 2020-06-11
Add  QnA meta tag to discussion

igraziatto:
Closes https://github.com/vanilla/addons/issues/816

### DIFF
--- a/plugins/QnA/class.qna.plugin.php
+++ b/plugins/QnA/class.qna.plugin.php
@@ -1041,9 +1041,9 @@ class QnAPlugin extends Gdn_Plugin {
     /**
      * Return QnA meta data tag string
      *
-     * @param stdClass $discussion
+     * @param object $discussion
      */
-    private function getDiscussionQnATagString(stdClass $discussion = null): string {
+    private function getDiscussionQnATagString(object $discussion = null): string {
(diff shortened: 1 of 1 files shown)


plugins/QnA/class.qna.plugin.php
```diff
@@ -1008,19 +1013,43 @@ public function discussionsController_unansweredCount_create($sender, $args) {
...
+            echo $this->getDiscussionQnATagString($discussion);
+        }
+    }
 
+    /**
+     * Return QnA meta data tag string
+     *
+     * @param stdClass $discussion
+     */
+    private function getDiscussionQnATagString(stdClass $discussion = null): string {
```
tburry, line 1046: Can you confirm that the discussion will always be a `stdClass`? If you are fairly confident I would at least type hint to `object`.

igraziatto: I've done a few tests with this coming from different controllers, it always comes as an object.

tburry: Okay just change the type hint to give us some flexibility for the future. Checking for stdClass specifically is rarely a good idea. It stopped us from making some wanted changes in our database layer before.

PR #820, opened 2020-06-18
Encode settings strings before outputting them

blutarch:
Closes vanilla/vanilla-patches#677

Two parameters that are set by the administrator are not being encoded, making them vulnerable to html injection. This PR runs the strings through the `htmlspecialchars()` method to prevent this.

### TO TEST
1. Enable the ShareThis plugin (make sure to set `hidden` to `false` in the plugin's `addon.json`)
1. In the plugin configuration settings, enter `aaaa"><script>alert(1337)</script>` in both the "Publisher Number" and "via Handle" fields.
1. Navigate to a discussion and note that the alert is triggered.
1. Checkout this branch and navigate to a discussion.
1. Verify that the alert is not triggered.

### DIFF
--- a/plugins/ShareThis/class.sharethis.plugin.php
+++ b/plugins/ShareThis/class.sharethis.plugin.php
@@ -12,24 +12,24 @@ class ShareThisPlugin extends Gdn_Plugin {
    /**
     * Show buttons after OP message body.
     */
-	public function discussionController_afterDiscussionBody_handler($sender) {
-      $publisherNumber = c('Plugin.ShareThis.PublisherNumber', 'Publisher Number');
-      $viaHandle = htmlspecialchars(c('Plugin.ShareThis.ViaHandle', ''));
-      $copyNShare = htmlspecialchars(c('Plugin.ShareThis.CopyNShare', false));
+    public function discussionController_afterDiscussionBody_handler($sender) {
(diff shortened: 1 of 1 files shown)


plugins/ShareThis/class.sharethis.plugin.php
```diff
@@ -14,8 +14,8 @@ class ShareThisPlugin extends Gdn_Plugin {
     */
 	public function discussionController_afterDiscussionBody_handler($sender) {
       $publisherNumber = c('Plugin.ShareThis.PublisherNumber', 'Publisher Number');
-      $viaHandle = c('Plugin.ShareThis.ViaHandle', '');
-      $copyNShare = c('Plugin.ShareThis.CopyNShare', false);
+      $viaHandle = htmlspecialchars(c('Plugin.ShareThis.ViaHandle', ''));
+      $copyNShare = htmlspecialchars(c('Plugin.ShareThis.CopyNShare', false));
```
charrondev, line 18:
```suggestion
      $copyNShare =c('Plugin.ShareThis.CopyNShare', false);
```

This is a boolean and isn't being outputted. My feeling is it will actually break the conditional check as well because it gets converted to a string through `htmlspecialchars`.

blutarch: @charrondev Thanks for catching this.

Issue #817, opened 2020-06-10
Sitemaps settings page enhancements


tburry:
Let's add an enhancement for the sitemaps.

**Change the wording for the settings description to:**

A site map is an xml file that search engines can use to help index your site. Learn more about sitemaps at [sitemaps.org]().

- Your main site map is located here: [http://dev.vanilla.localhost/sitemapindex.xml](#)
- Your [robots.txt](#) file contains the location of this site map so that search engines that are aware of your site will know where to look.

**Add a toggle to the settings page to enable the feature flag**

**Label**: Discussion Based Sitemaps (BETA)
**Description**: Use the sitemaps that point directly to discussions instead of categories.

PR #821, opened 2020-06-19
Add discussionSiteMaps toggle, and fix plugin description

dannimajdalani:
closes https://github.com/vanilla/addons/issues/817

**To Test**

- Check the description in the referenced ticket.
- Test the toggle, it should set ['Feature']['discussionSiteMaps']['Enabled']
- Enable private community, check the settings page description.

### DIFF
--- a/plugins/Sitemaps/class.sitemaps.plugin.php
+++ b/plugins/Sitemaps/class.sitemaps.plugin.php
@@ -187,7 +187,7 @@ class SitemapsPlugin extends Gdn_Plugin {
         $configurationModule = new ConfigurationModule($sender);
         $configurationModule->initialize([
             'Feature.discussionSiteMaps.Enabled' => [
-                'LabelCode' => t('Discussion Based Sitemaps (BETA)'),
+                'LabelCode' => t('Discussion Based Sitemaps', 'Discussion Based Sitemaps (BETA)'),
                 'Control' => 'Toggle',
                 'Description' => t('Use the sitemaps that point directly to discussions instead of categories.'),
             ]]);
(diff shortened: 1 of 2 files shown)


plugins/Sitemaps/class.sitemaps.plugin.php
```diff
@@ -183,6 +183,15 @@ public function settingsController_sitemaps_create($sender) {
         $sender->setData('Title', t('Sitemap Settings'));
         $sender->setData('isSitePrivate', $this->isSitePrivate);
         $sender->addSideMenu();
+
+        $configurationModule = new ConfigurationModule($sender);
+        $configurationModule->initialize([
+            'Feature.discussionSiteMaps.Enabled' => [
+                'LabelCode' => t('Discussion Based Sitemaps (BETA)'),
```
tburry, line 190:
```suggestion
                'LabelCode' => t('Discussion Based Sitemaps', 'Discussion Based Sitemaps (BETA)'),
```

This way the translation code won't change when this goes out of beta. This might also be a case where concatenation is acceptable because then we wouldn't have to retranslate either.

plugins/Sitemaps/views/settings.twig
```diff
@@ -1,14 +1,13 @@
 <h1>{{Title}}</h1>
-<div class="padded">
+<div class="padded content">
     {% if isSitePrivate %}
-    <p class="Message AlertMessage">
-        Private Communities is currently enabled. Sitemaps are not recommended with Private Communities unless you know what you're doing.
-    </p>
+       <p class="Message AlertMessage">
+         Private Communities is currently enabled. Sitemaps are not recommended with Private Communities unless you know what you're doing.
```
tburry, line 5:
```suggestion
         Private communities are currently enabled. Sitemaps are not recommended with private communities unless you know what you're doing.
```

PR #831, opened 2020-07-09
Remove deleted user from ignore lists

blutarch:
Closes vanilla/support#2241

Currently, when a user is deleted, if that user is on anyone's ignore list, their name changes to "[Deleted User]" and they can no longer be removed from the list because there's no longer a user in the db. This PR fixes that problem by removing a user from all ignore lists before deleting the user.

### TO TEST
1. As an admin, create a user and ignore the user.
1. Delete the user and view your ignore list.
1. Note that you now have "[Deleted User]" on your ignore list and cannot unignore them.
1. Check out this branch and repeat steps 1 and 2.
1. Note that neither the user nor "[Deleted User]" is on your ignore list.

### DIFF
--- a/plugins/Ignore/class.ignore.plugin.php
+++ b/plugins/Ignore/class.ignore.plugin.php
@@ -713,7 +713,7 @@ class IgnorePlugin extends Gdn_Plugin {
    }
 
     /**
-     * If a user is being deleting, remove them from other users' ignore lists.
+     * If a user is being deleted, remove them from other users' ignore lists.
      *
      * @param UserController $sender
      */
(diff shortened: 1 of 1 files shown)


nervoustwit:
Is there not a way to, when you query the user table for ignored users, that you put a clause in there that says do not return deleted users? Because if I was already ignoring someone before this PR they remain on my list. It seems like that would be easier.

Right here: https://github.com/vanilla/addons/blob/master/plugins/Ignore/class.ignore.plugin.php#L168

if you do something like:
```
foreach ($ignoredUsers as $ignoredUsersID => &$ignoredUser) {
   if (!$ignoredUser[$ignoredUsersID]['Deleted']) {
         $ignoredUser['IgnoreDate'] = $ignoredUsersIDs[$ignoredUsersID];
   }
}
```

blutarch:
I'm closing this, as we are going to handle this problem in a different way.

PR #838, opened 2020-11-25
Fix checking for troll users before comment save event

dannimajdalani:
related https://github.com/vanilla/vanilla-cloud/pull/1558


blutarch (approved):
Tested locally with discussions and comments. Also tested in a nested category. Works perfectly.

igraziatto (approved):
Tested locally, it works as expected.

PR #844, opened 2021-03-01
Add ban by Fingerprint to the Troll management plugin

vanilla-dbarbier:
This is the 'plugin' part for the following issue: https://github.com/vanillaforums/mse/issues/236
As detailed by the following estimate: https://github.com/vanillaforums/estimates/blob/master/mse/mse-troll-plugin-enhancements.md

This is complementary to the following PR: https://github.com/vanilla/vanilla-cloud/pull/2303

**How to test**

- Enable the **Troll management plugin**.
- Configure the plugin.
- Enable fingerprint checks.
- Set a maximum allowed number of user accounts tied to a single fingerprint.
- Logout from the admin account.
- Create a few new user accounts to reach the maximum allowed number of user accounts tied to a single fingerprint.
- Logout & back again as your administrative account.
- Reach the applicants management interface(/dashboard/user/applicants). Your newest user should be listed.

**Notes**

- Added configurations & corresponding screens to allow modification of the amount accounts sharing the same fingerprint.
- A few tweaks were done to the _fingerprinting_ process.
- The user registration process verifies if the number of allowed accounts using the same fingerprint is reached. If it is the case, the newly registered user is set as _applicant_.

### DIFF
--- a/plugins/TrollManagement/TrollManagementTest.php
+++ b/plugins/TrollManagement/TrollManagementTest.php
@@ -1,0 +1,71 @@
+<?php
+/**
+ * @author David Barbier<david.barbier@vanillaforums.com>
+ * @copyright 2009-2021 Vanilla Forums Inc.
+ * @license GPL-2.0-only
+ */
+
+namespace VanillaTests\TrollManagement;
+
+use UserModel;
+use RoleModel;
+use VanillaTests\Models\UserModelTest;
+use VanillaTests\SetupTraitsTrait;
+
+/**
+ * Class TrollManagementTest
+ */
+class TrollManagementTest extends UserModelTest {
+    use SetupTraitsTrait;
+
+    /** @var UserModel */
+    protected $userModel;
+
+    /**
+     * @inheritDoc
+     */
+    public function setUp(): void {
+        parent::setUp();
+    }
+
+    /**
+     * Setup routine, run before the test class is instantiated.
+     */
+    public static function setupBeforeClass(): void {
+        self::$addons = ['vanilla', 'trollmanagement'];
+
+        parent::setupBeforeClass();
+    }
+
+    /**
+     * Tests the automatic assignment of the "applicant" status to a new user registration if the maximum amount of
+     * user accounts sharing the same fingerprint is reached.
+     */
+    public function testRegisterApplicant(): void {
+        /** @var \Gdn_Configuration $configuration */
+        $configuration = static::container()->get('Config');
+
+        $configuration->set('TrollManagement.PerFingerPrint.Enabled', true);
+        $configuration->set('TrollManagement.PerFingerPrint.MaxUserAccounts', 3);
+
+        // Ensure all future registered dummy users uses the same fingerprint.
+        $_COOKIE['__vnf'] = 'THISISAFAKEFINGERPRINT';
+
+        // Create 3 dummy accounts. (The third should be automatically set as "applicant")
+        $importedUsers[] = $this->insertDummyUser();
+        $importedUsers[] = $this->insertDummyUser();
+        $importedUsers[] = $this->insertDummyUser();
+
+        // We pull the associated user's roles.
+        foreach ($importedUsers as $importedUser) {
+            $importedUsersRolesIDs[] = $this->userModel->getRoleIDs($importedUser['UserID'])/*->resultArray()*/;
+        }
+
+        // The FIRST dummy user account is NOT an applicant
+        $this->assertNotContains(RoleModel::APPLICANT_ID, $importedUsersRolesIDs['0']);
+        // The SECOND dummy user account is NOT an applicant
+        $this->assertNotContains(RoleModel::APPLICANT_ID, $importedUsersRolesIDs['1']);
+        // The THIRD dummy user account is NOT an applicant
+        $this->assertContains(RoleModel::APPLICANT_ID, $importedUsersRolesIDs['2']);
+     }
+}
--- a/plugins/TrollManagement/TrollManagementTest.php
+++ b/plugins/TrollManagement/TrollManagementTest.php
@@ -58,7 +58,7 @@ class TrollManagementTest extends UserModelTest {
 
         // We pull the associated user's roles.
         foreach ($importedUsers as $importedUser) {
-            $importedUsersRolesIDs[] = $this->userModel->getRoleIDs($importedUser['UserID'])/*->resultArray()*/;
+            $importedUsersRolesIDs[] = $this->userModel->getRoleIDs($importedUser['UserID']);
         }
 
         // The FIRST dummy user account is NOT an applicant
@@ -67,5 +67,54 @@ class TrollManagementTest extends UserModelTest {
         $this->assertNotContains(RoleModel::APPLICANT_ID, $importedUsersRolesIDs['1']);
         // The THIRD dummy user account is NOT an applicant
         $this->assertContains(RoleModel::APPLICANT_ID, $importedUsersRolesIDs['2']);
+    }
+
+    /**
+     * Tests the automatic assignment of the "applicant" status to EVERY new user registration.
+     */
+    public function testRegisterAllApplicants(): void {
+        /** @var \Gdn_Configuration $configuration */
+        $configuration = static::container()->get('Config');
+
+        $configuration->set('TrollManagement.PerFingerPrint.Enabled', true);
+        $configuration->set('TrollManagement.PerFingerPrint.MaxUserAccounts', 0);
+
+        // Create 3 dummy accounts. (They should all be automatically set as "applicant")
+        $importedUsers[] = $this->insertDummyUser();
+        $importedUsers[] = $this->insertDummyUser();
+        $importedUsers[] = $this->insertDummyUser();
+
+        // We pull the associated user's roles.
+        foreach ($importedUsers as $importedUser) {
+            $importedUsersRolesIDs = $this->userModel->getRoleIDs($importedUser['UserID']);
+            // The dummy user account is an applicant.
+            $this->assertContains(RoleModel::APPLICANT_ID, $importedUsersRolesIDs);
+        }
+    }
+
+    /**
+     * The plugin is disabled. Even if every users are using the same fingerprint, none are flagged as "applicant".
+     */
+    public function testRegisterDisabledPlugin(): void {
+        /** @var \Gdn_Configuration $configuration */
+        $configuration = static::container()->get('Config');
+
+        $configuration->set('TrollManagement.PerFingerPrint.Enabled', false);
+        $configuration->set('TrollManagement.PerFingerPrint.MaxUserAccounts', 1);
+
+        // Ensure all future registered dummy users uses the same fingerprint.
+        $_COOKIE['__vnf'] = 'THISISAFAKEFINGERPRINT';
+
+        // Create 3 dummy accounts. (None should be automatically set as "applicant")
+        $importedUsers[] = $this->insertDummyUser();
+        $importedUsers[] = $this->insertDummyUser();
+        $importedUsers[] = $this->insertDummyUser();
+
+        // We pull the associated user's roles.
+        foreach ($importedUsers as $importedUser) {
+            $importedUsersRolesIDs = $this->userModel->getRoleIDs($importedUser['UserID']);
+            // The dummy user account is NOT an applicant
+            $this->assertNotContains(RoleModel::APPLICANT_ID, $importedUsersRolesIDs);
+        }
     }
 }
--- a/plugins/TrollManagement/TrollManagementTest.php
+++ b/plugins/TrollManagement/TrollManagementTest.php
@@ -67,5 +67,5 @@ class TrollManagementTest extends UserModelTest {
         $this->assertNotContains(RoleModel::APPLICANT_ID, $importedUsersRolesIDs['1']);
         // The THIRD dummy user account is NOT an applicant
         $this->assertContains(RoleModel::APPLICANT_ID, $importedUsersRolesIDs['2']);
-     }
+    }
 }
--- a/plugins/TrollManagement/class.trollmanagement.plugin.php
+++ b/plugins/TrollManagement/class.trollmanagement.plugin.php
@@ -35,11 +35,22 @@ class TrollManagementPlugin extends Gdn_Plugin {
     private $session;
 
     /**
+     * @var UserModel
+     */
+    private $userModel;
+
+    /**
      * TrollManagementPlugin constructor.
      *
      * @param Gdn_Session $session Injected session.
      */
-    public function __construct(Gdn_Session $session) {
+    public function __construct(Gdn_Session $session, UserModel $userModel = null) {
+        if ($userModel === null) {
+            $this->userModel = Gdn::getContainer()->get(UserModel::class);
+        } else {
+            $this->userModel = $userModel;
+        }
+
         parent::__construct();
         $this->session = $session;
     }
@@ -578,15 +589,12 @@ class TrollManagementPlugin extends Gdn_Plugin {
             return false;
         }
 
-        $sql = clone Gdn::sql();
-        $sql->reset();
-        $users = $sql
+        $usersCount = $this->userModel->SQL
             ->select('userID')
             ->from('User')
             ->where('Fingerprint', $fingerprint)
-            ->limit($maxSiblingAccounts)
-            ->get()->resultArray();
-        return (count($users) == $maxSiblingAccounts);
+            ->limit($maxSiblingAccounts)->get()->numRows();
+        return ($usersCount == $maxSiblingAccounts);
     }
 
     /**
--- a/plugins/TrollManagement/class.trollmanagement.plugin.php
+++ b/plugins/TrollManagement/class.trollmanagement.plugin.php
@@ -76,9 +76,6 @@ class TrollManagementPlugin extends Gdn_Plugin {
      * On plugin disable.
      */
     public function onDisable() {
-        // TODO: Ask Ryan about best practices here.
-        // TODO: Ban records that uses the BanType "Fingerprint"
-        // TODO: Remove "Fingerprint" from the table's Bantype enum.
     }
 
     /**
--- a/plugins/TrollManagement/class.trollmanagement.plugin.php
+++ b/plugins/TrollManagement/class.trollmanagement.plugin.php
@@ -75,12 +75,6 @@ class TrollManagementPlugin extends Gdn_Plugin {
     }
 
     /**
-     * On plugin disable.
-     */
-    public function onDisable() {
-    }
-
-    /**
      * Get list of current troll user IDs.
      *
      * @return array
@@ -559,11 +553,9 @@ class TrollManagementPlugin extends Gdn_Plugin {
             $maxSiblingAccounts = c('TrollManagement.PerFingerPrint.MaxUserAccounts');
             $userFingerprint = val('Fingerprint', $args['User']);
             if (!empty($userFingerprint)) {
-                $fingerprintUsages = $this->getSharedFingerprintsUsersCount($userFingerprint);
-                if ($fingerprintUsages >= $maxSiblingAccounts) {
+                if ($this->checkMaxSharedFingerprintsExceeded($userFingerprint, $maxSiblingAccounts)) {
                     $sender->EventArguments['ApplicantMeta'][t("Fingerprint issue")] = sprintf(
-                        t("%s accounts are sharing the '%s' fingerprint."),
-                        $fingerprintUsages,
+                        t("Too many accounts are using the '%s' fingerprint."),
                         $userFingerprint
                     );
                 }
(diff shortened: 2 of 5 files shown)


plugins/TrollManagement/class.trollmanagement.plugin.php
```diff
@@ -469,4 +499,179 @@ private function checkTroll(int $userID, array &$args) {
...
+     * accounts using the same fingerprint.
+     *
+     * @param Controller $sender
+     * @param array $args
+     */
+    public function base_applicantInfo_handler($sender, $args) {
+        if (c('TrollManagement.PerFingerPrint.Enabled', false)) {
+            $maxSiblingAccounts = c('TrollManagement.PerFingerPrint.MaxUserAccounts');
+            $userFingerprint = val('Fingerprint', $args['User']);
+            $fingerprintUsages = $this->getSharedFingerprintsUsersCount($userFingerprint);
```
blutarch, line 561: What happens if `$userFingerprint` is `null` and you call `getSharedFingerprintsUsersCounts()`?

plugins/TrollManagement/class.trollmanagement.plugin.php
```diff
@@ -469,4 +499,179 @@ private function checkTroll(int $userID, array &$args) {
...
+     * @param UserModel $sender
+     * @param array $args
+     */
+    public function userModel_afterRegister_handler($sender, $args) {
+        $userID = $args['UserID'];
+
+        $maxSiblingAccounts = c('TrollManagement.PerFingerPrint.MaxUserAccounts');
+        $userFingerprint = $this->setFingerprint($userID);
+
+        $fingerprintUsages = $this->getSharedFingerprintsUsersCount($userFingerprint);
```
blutarch, line 602: I think you need to handle the case where `setFingerpring()` returns false.

blutarch (changes requested):
See the two comments I left.

vanilla-dbarbier:
> Make sure we get some tests in here.

3 tests were added.

blutarch (approved):
Tested locally.

plugins/TrollManagement/TrollManagementTest.php
```diff
@@ -0,0 +1,120 @@
+<?php
```
initvector, line 1: Tests should go in a "tests" directory under the plugin. Otherwise, our test runner won't find them.

plugins/TrollManagement/TrollManagementTest.php
```diff
@@ -0,0 +1,120 @@
...
+
+    /** @var UserModel */
+    protected $userModel;
+
+    /**
+     * @inheritDoc
+     */
+    public function setUp(): void {
+        parent::setUp();
+    }
```
initvector, line 29: What is this doing?

plugins/TrollManagement/TrollManagementTest.php
```diff
@@ -0,0 +1,120 @@
...
+
+use UserModel;
+use RoleModel;
+use VanillaTests\Models\UserModelTest;
+use VanillaTests\SetupTraitsTrait;
+
+/**
+ * Class TrollManagementTest
+ */
+class TrollManagementTest extends UserModelTest {
```
initvector, line 18: I don't think you want to extend `UserModelTest`. You're going to be running every test in that base class too.

plugins/TrollManagement/TrollManagementTest.php
```diff
@@ -0,0 +1,120 @@
...
+        // We pull the associated user's roles.
+        foreach ($importedUsers as $importedUser) {
+            $importedUsersRolesIDs[] = $this->userModel->getRoleIDs($importedUser['UserID']);
+        }
+
+        // The FIRST dummy user account is NOT an applicant
+        $this->assertNotContains(RoleModel::APPLICANT_ID, $importedUsersRolesIDs['0']);
+        // The SECOND dummy user account is NOT an applicant
+        $this->assertNotContains(RoleModel::APPLICANT_ID, $importedUsersRolesIDs['1']);
+        // The THIRD dummy user account is NOT an applicant
```
initvector, line 68: > // The THIRD dummy user account is NOT an applicant

🤔

plugins/TrollManagement/TrollManagementTest.php
```diff
@@ -0,0 +1,120 @@
...
+        $this->assertContains(RoleModel::APPLICANT_ID, $importedUsersRolesIDs['2']);
+    }
+
+    /**
+     * Tests the automatic assignment of the "applicant" status to EVERY new user registration.
+     */
+    public function testRegisterAllApplicants(): void {
```
initvector, line 75: This seems a little odd. If we wanted everyone to be an applicant, I'd think we would just enable applicant registration.

plugins/TrollManagement/TrollManagementTest.php
initvector, line 18: Should really add some "banning" tests in here, since that's a large part of what this PR is introducing.

plugins/TrollManagement/TrollManagementTest.php
```diff
@@ -0,0 +1,120 @@
...
+        // We pull the associated user's roles.
+        foreach ($importedUsers as $importedUser) {
+            $importedUsersRolesIDs = $this->userModel->getRoleIDs($importedUser['UserID']);
+            // The dummy user account is an applicant.
+            $this->assertContains(RoleModel::APPLICANT_ID, $importedUsersRolesIDs);
+        }
+    }
+
+    /**
+     * The plugin is disabled. Even if every users are using the same fingerprint, none are flagged as "applicant".
```
initvector, line 96: The plug-in is still technically enabled, right? Just not this "per-fingerprint" feature?

plugins/TrollManagement/class.trollmanagement.plugin.php
```diff
@@ -55,13 +55,31 @@ public function setup() {
...
             ->column('Troll', 'int', '0')
             ->column('Fingerprint', 'varchar(50)', null, 'index')
             ->set();
     }
 
+    /**
+     * On plugin disable.
+     */
+    public function onDisable() {
+    }
```
initvector, line 81: 🤔

plugins/TrollManagement/class.trollmanagement.plugin.php
```diff
@@ -469,4 +499,189 @@ private function checkTroll(int $userID, array &$args) {
...
+     * Add "Fingerprint" to the possible ban query.
+     *
+     * @param array $result
+     * @param array $ban
+     * @return array
+     */
+    public function banModel_banWhere_handler(array $result, array $ban): array {
+        switch (strtolower($ban['BanType'])) {
+            case 'fingerprint':
+                $result['u.Fingerprint like'] = $ban['BanValue'];
```
initvector, line 659: I don't think a "like" is useful here. Aren't these values more or less random? Why would you want to ban based on a partial match?

plugins/TrollManagement/class.trollmanagement.plugin.php
```diff
@@ -469,4 +499,189 @@ private function checkTroll(int $userID, array &$args) {
...
+        }
+    }
+
+    /**
+     * Return a count of users using the same provided fingerprint.
+     *
+     * @param null|string $fingerprint
+     * @return int
+     */
+    public function getSharedFingerprintsUsersCount($fingerprint): int {
```
initvector, line 580: Does this need to be public?

plugins/TrollManagement/class.trollmanagement.plugin.php
```diff
@@ -469,4 +499,189 @@ private function checkTroll(int $userID, array &$args) {
...
+        if (!is_null($fingerprint)) {
+            $sql = clone Gdn::sql();
+            $sql->reset();
+            $users = $sql
+                ->select('userID AS siblingsCount', 'count')
```
initvector, line 585: As often as this query has the potential to be run, trying to perform a count each time isn't going to be so performant. It looks like you're only ever comparing this against the "max siblings". Maybe just query the user table for rows with that fingerprint, limiting by the "max siblings" value. If you have at least that many, we know this fingerprint has exceeded its account allowance.

initvector (changes requested):

plugins/TrollManagement/TrollManagementTest.php
vanilla-dbarbier: Gotcha.

plugins/TrollManagement/TrollManagementTest.php
vanilla-dbarbier: Previously I had dummy user accounts created there but as the Troll plugin needs to be configured prior to the user creation, I moved it. At the moment, this is effectively doing nothing much. Removing it.

plugins/TrollManagement/TrollManagementTest.php
vanilla-dbarbier: Oops. Obvious cut & paste mishap. Changed it.

plugins/TrollManagement/TrollManagementTest.php
vanilla-dbarbier: Right, it's the feature. :|

plugins/TrollManagement/class.trollmanagement.plugin.php
vanilla-dbarbier: Remnants of me wanting to cleanup after disabling the plugin. Removed.

plugins/TrollManagement/TrollManagementTest.php
vanilla-dbarbier: Right. Replaced it with SiteTestCase.

plugins/TrollManagement/class.trollmanagement.plugin.php
vanilla-dbarbier: Not at all. Changing it.

plugins/TrollManagement/class.trollmanagement.plugin.php
vanilla-dbarbier: Alright. This has been adjusted.

plugins/TrollManagement/TrollManagementTest.php
vanilla-dbarbier: Agreed. Should I just remove the test or only allow values starting at **1**?

plugins/TrollManagement/TrollManagementTest.php
vanilla-dbarbier: Agreed.
One test has been added: a simple ban per fingerprint.
Trying to think of another plausible usage...

plugins/TrollManagement/class.trollmanagement.plugin.php
```diff
@@ -469,4 +493,191 @@ private function checkTroll(int $userID, array &$args) {
...
+     *
+     * @param null|string $fingerprint
+     * @param int $maxSiblingAccounts
+     * @return bool
+     */
+    private function checkMaxSharedFingerprintsExceeded($fingerprint, int $maxSiblingAccounts): bool {
+        if (!is_null($fingerprint)) {
+            $sql = clone Gdn::sql();
+            $sql->reset();
+            $users = $sql
```
initvector, line 577: Do we need to use the naked SQL object? It's probably better to use the user model, if possible.

plugins/TrollManagement/class.trollmanagement.plugin.php
```diff
@@ -469,4 +493,191 @@ private function checkTroll(int $userID, array &$args) {
...
+                ->select('userID')
+                ->from('User')
+                ->where('Fingerprint', $fingerprint)
+                ->limit($maxSiblingAccounts)
+                ->get()->resultArray();
+                //firstRow(DATASET_TYPE_ARRAY);
```
initvector, line 583:
```suggestion
```

plugins/TrollManagement/class.trollmanagement.plugin.php
```diff
@@ -469,4 +493,191 @@ private function checkTroll(int $userID, array &$args) {
...
+    private function checkMaxSharedFingerprintsExceeded($fingerprint, int $maxSiblingAccounts): bool {
+        if (!is_null($fingerprint)) {
+            $sql = clone Gdn::sql();
+            $sql->reset();
+            $users = $sql
+                ->select('userID')
+                ->from('User')
+                ->where('Fingerprint', $fingerprint)
+                ->limit($maxSiblingAccounts)
+                ->get()->resultArray();
```
initvector, line 582: There's a `numRows` method on the `Gdn_Dataset` object that will probably be better here than `resultArray`.

plugins/TrollManagement/tests/TrollManagementTest.php
```diff
@@ -0,0 +1,153 @@
...
+        // The SECOND dummy user account is NOT an applicant
+        $this->assertNotContains(RoleModel::APPLICANT_ID, $importedUsersRolesIDs['1']);
+        // The THIRD dummy user account should an applicant
+        $this->assertContains(RoleModel::APPLICANT_ID, $importedUsersRolesIDs['2']);
+    }
+
+    /**
+     * Tests the automatic assignment of the "applicant" status to EVERY new user registration.
+     */
+    public function testRegisterAllApplicants(): void {
```
initvector, line 72: If we're adding a test here, we're essentially saying this is expected and intended behavior. Is adding "make every new registration an applicant" an intended behavior of the Troll Management addon? As discussed, we already have the applicant registration method to make all new registrations come in as applications. Instead, I would recommend enforcing "MaxUserAccounts" be no less than 1.

initvector (changes requested):

plugins/TrollManagement/class.trollmanagement.plugin.php
```diff
@@ -469,4 +493,191 @@ private function checkTroll(int $userID, array &$args) {
...
+
+    /**
+     * Checks if the maximum amount of accounts using the same fingerprint has been reached.
+     *
+     * @param null|string $fingerprint
+     * @param int $maxSiblingAccounts
+     * @return bool
+     */
+    private function checkMaxSharedFingerprintsExceeded($fingerprint, int $maxSiblingAccounts): bool {
+        if (!is_null($fingerprint)) {
```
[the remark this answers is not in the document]
vanilla-dbarbier: Alright. Corrected.

plugins/TrollManagement/class.trollmanagement.plugin.php
vanilla-dbarbier: Removed.

plugins/TrollManagement/TrollManagementTest.php
vanilla-dbarbier: I removed the test & added a bessy test that checks for validation in the backend.

plugins/TrollManagement/tests/TrollManagementTest.php
vanilla-dbarbier: Agreed. I removed the test & added a bessy test that checks for validation in the backend.

plugins/TrollManagement/class.trollmanagement.plugin.php
vanilla-dbarbier: Alright-o. Done.

plugins/TrollManagement/class.trollmanagement.plugin.php
vanilla-dbarbier: Done.

plugins/TrollManagement/class.trollmanagement.plugin.php
```diff
@@ -469,4 +493,191 @@ private function checkTroll(int $userID, array &$args) {
...
+    }
+
+    /**
+     * Checks if the maximum amount of accounts using the same fingerprint has been reached.
+     *
+     * @param null|string $fingerprint
+     * @param int $maxSiblingAccounts
+     * @return bool
+     */
+    private function checkMaxSharedFingerprintsExceeded($fingerprint, int $maxSiblingAccounts): bool {
```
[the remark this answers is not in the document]
vanilla-dbarbier: Yes.

vanilla-dbarbier:
@initvector just so I have a clear conscience about it, this is a reminder that this PR too is pending review.

plugins/TrollManagement/class.trollmanagement.plugin.php
```diff
@@ -469,4 +505,201 @@ private function checkTroll(int $userID, array &$args) {
...
+    }
+
+    /**
+     * Checks if the maximum amount of accounts using the same fingerprint has been reached.
+     *
+     * @param string $fingerprint
+     * @param int $maxSiblingAccounts
+     * @return bool
+     */
+    private function checkMaxSharedFingerprintsExceeded(string $fingerprint, int $maxSiblingAccounts): bool {
```
initvector, line 588: We added a type to `$fingerprint`, which is great`, but if `null` is also a valid value, we need to specify that in the type declaration.

```suggestion
    private function checkMaxSharedFingerprintsExceeded(?string $fingerprint, int $maxSiblingAccounts): bool {
```

plugins/TrollManagement/class.trollmanagement.plugin.php
```diff
@@ -469,4 +505,201 @@ private function checkTroll(int $userID, array &$args) {
...
+        if (is_null($fingerprint)) {
+            return false;
+        }
+
+        $usersCount = $this->userModel->SQL
```
initvector, line 593: It looks like you replaced using the global SQL object with using the one in `UserModel`. But couldn't you just use the methods on the model used for querying users, like `getWhere`?

plugins/TrollManagement/class.trollmanagement.plugin.php
```diff
@@ -469,4 +505,201 @@ private function checkTroll(int $userID, array &$args) {
...
+if (!function_exists('validatePositiveNumber')) {
+    /**
+     * Validate if number is positive
+     *
+     * @param int|string $number
+     * @return bool
+     */
+    function validatePositiveNumber($number): bool {
+        return (is_numeric($number) && (int)$number > 0);
+    }
```
initvector, line 704: Was this copied & pasted from the QnA addon? If it's needed in multiple places, it's something we should be adding to core.

plugins/TrollManagement/tests/TrollManagementTest.php
```diff
@@ -0,0 +1,188 @@
...
+    public function setUp(): void {
+        parent::setUp();
+        $this->createUserFixtures();
+    }
+
+    /**
+     * Setup routine, run before the test class is instantiated.
+     */
+    public static function setupBeforeClass(): void {
+        self::$addons = ['vanilla', 'trollmanagement'];
```
initvector, line 40: Do we need "vanilla" here? If not, removing it and avoiding its setup would be beneficial.

plugins/TrollManagement/tests/TrollManagementTest.php
```diff
@@ -0,0 +1,188 @@
...
+     */
+    public function testRegisterApplicant(): void {
+        /** @var \Gdn_Configuration $configuration */
+        $configuration = static::container()->get('Config');
+
+        $configuration->set('TrollManagement.PerFingerPrint.Enabled', true);
+        $configuration->set('TrollManagement.PerFingerPrint.MaxUserAccounts', 3);
+
+        // Ensure all future registered dummy users uses the same fingerprint.
+        $_COOKIE['__vnf'] = 'THISISAFAKEFINGERPRINT';
```
initvector, line 57: For the sake of tidying up the environment, this should probably be unset as part of the test case's `teardown`.

plugins/TrollManagement/tests/TrollManagementTest.php
```diff
@@ -0,0 +1,188 @@
...
+        // We pull the associated user's roles.
+        foreach ($importedUsers as $importedUser) {
+            $importedUsersRolesIDs[] = $this->userModel->getRoleIDs($importedUser['UserID']);
+        }
+
+        // The FIRST dummy user account is NOT an applicant
+        $this->assertNotContains(RoleModel::APPLICANT_ID, $importedUsersRolesIDs['0']);
+        // The SECOND dummy user account is NOT an applicant
+        $this->assertNotContains(RoleModel::APPLICANT_ID, $importedUsersRolesIDs['1']);
+        // The THIRD dummy user account should an applicant
```
initvector, line 73: This setting is documented in the UI as "maximum allowed number of user accounts tied to a single fingerprint", but if I say the maximum number of accounts a user can have is three, I would expect them to be able to register three accounts. If I'm reading this correctly, we're actually only allowing them to register two accounts.

plugins/TrollManagement/class.trollmanagement.plugin.php
```diff
@@ -469,4 +505,201 @@ private function checkTroll(int $userID, array &$args) {
...
+
+        $sender->title(sprintf(t('%s settings'), t('Troll Management')));
+        $sender->setData('PluginDescription', $this->getPluginKey('Description'));
+
+        $sender->Form = new Gdn_Form();
+        $validation = new Gdn_Validation();
+        $configurationModel = new Gdn_ConfigurationModel($validation);
+
+        $configurationModel->setField([
+            'TrollManagement.PerFingerPrint.Enabled' => c('TrollManagement.PerFingerPrint.Enabled', false),
```
initvector, line 526: This is a little nit-picky, but let's keep the casing of "fingerprint" consistent (e.g. `Fingerprint` versus `FingerPrint`).

initvector (changes requested):
Getting close! Few minor comments here. We should also be looking toward adding translations for any new strings.

plugins/TrollManagement/tests/TrollManagementTest.php
vanilla-dbarbier: Agreed. The test has been adjusted.

plugins/TrollManagement/class.trollmanagement.plugin.php
vanilla-dbarbier: Yes. I modified it.

plugins/TrollManagement/tests/TrollManagementTest.php
vanilla-dbarbier: Agreed. It is faster.
Done.

plugins/TrollManagement/tests/TrollManagementTest.php
vanilla-dbarbier: Agreed. Done.

plugins/TrollManagement/class.trollmanagement.plugin.php
vanilla-dbarbier: Agreed. Moved it to functions.validation.php

initvector (approved):
🤞

PR #846, opened 2021-05-03
Fix typos preventing Troll Management plugin from saving.

dlacaille:
Closes https://github.com/vanilla/support/issues/4167

This fixes troll management plugin not saving properly because of typos in the settings keys.


blutarch (approved):
Tested locally. Works as expected.

PR #862, opened 2021-12-13
VNLA-848: Fix jQuery being called early in the stopAutoDraft plugin

ManeeshChiba:
## Changes
- Wraps the existing code in `onVanillaReady`

### Steps to replicate
- Checkout master
- Turn on the Stop AutoDraft plugin
- Go to a new discussion page
- Observe that `jQuery is undefined` in the developer console
- Switch to this branch
- Reload the page
- Observe the error about `jQuery` is gone

Discovered during investigating https://higherlogic.atlassian.net/browse/VNLA-848


ManeeshChiba:
This should plugin needs to be removed and the stopping auto draft turned into a config setting.

PR #864, opened 2022-01-28
Fix/appeng 13922/add depricated warning to share this plugin

mgriffhl:
Request from support to add deprecation on ShareThis plugin - plugin was still showing on staging but not on production.

Closes: https://higherlogic.atlassian.net/browse/APPENG-13922


mgriffhl:
Meant to merge to master

PR #871, opened 2022-09-02
VNLA-1083 Fix Link Types working with warn on leave

mike-masson:
This fixes sites that are using the Warn on Leave feature with the LinkTypes plugin as the warn links would not open in a new tab, previously. 

The regex is simple and efficient with an almost impossible chance to get a false positive.

Tested locally on a standard build and using basic embed. Was not able to test on the advanced embed due to CORS errors on two test sites.


mike-masson:
@charrondev As far as I can tell those tests should fail on the addons repo; could you confirm?

andrew-higherlogic (approved):
Looks good.

PR #872, opened 2023-01-13
VNLA-1121: Disable StopAutoDraft Plugin Universally

blutarch:
Companion PR to vanilla/vanilla-cloud#5559

This PR changes the logic of the StopAutoDraft plugin to simply set the new `Vanilla.Drafts.Autosave` config value to `false`. This setting is used by the new settings option on the `vanilla/settings/posting` page. Once the config value is set, the addon disables itself. The PR also hides the plugin, so it cannot be found subsequently.

### DIFF
--- a/plugins/StopAutoDraft/class.stopautodraft.plugin.php
+++ b/plugins/StopAutoDraft/class.stopautodraft.plugin.php
@@ -23,6 +23,13 @@ class StopAutoDraftPlugin extends Gdn_Plugin
     /** @var AddonManager  */
     protected $addonManager;
 
+    /**
+     * D.I.
+     *
+     * @param \Vanilla\Contracts\ConfigurationInterface $config
+     * @param \Vanilla\Models\AddonModel $addonModel
+     * @param AddonManager $addonManager
+     */
     public function __construct(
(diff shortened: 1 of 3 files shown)


plugins/StopAutoDraft/class.stopautodraft.plugin.php
```diff
@@ -8,28 +12,36 @@
...
+        $this->addonModel = $addonModel;
+        $this->addonManager = $addonManager;
+        parent::__construct();
+    }
+
+    public function container_init()
+    {
+        $this->config->saveToConfig("Vanilla.Drafts.Autosave", false);
+        $addon = $this->addonManager->lookupAddon("stopautodraft");
+        $this->addonModel->disable($addon);
```
charrondev, line 41: This should definitely not be done in container init. I'd recommend sticking it in structure.

charrondev (changes requested):

plugins/StopAutoDraft/class.stopautodraft.plugin.php
blutarch: Done

charrondev (approved):

